PR #111, opened 2020-04-30
Add additional Elixir eex languages

zolrath:
Hello!
So there's an interesting amount of language ids involved in supporting HTML (Eex) for `html.eex` and `html.leex` files.
The original Language ID was `HTML (Eex)` which was then replaced with `HTML (EEx)` in a future plugin.

New Identifier Guidelines for language ids specify lowercase identifiers with no spaces.
https://code.visualstudio.com/docs/languages/identifiers#_new-identifier-guidelines

This will hopefully be corrected with:
https://github.com/elixir-lsp/vscode-elixir-ls/pull/87

By supporting `HTML (Eex)` and `HTML (EEx)` you can support both the legacy Elixir plugin as well as the current language service plugin.

By supporting `html-eex` you will avoid the plugin suddenly not working when the language id change in the new plugin takes place.

### DIFF
--- a/packages/tailwindcss-language-server/src/util/html.ts
+++ b/packages/tailwindcss-language-server/src/util/html.ts
@@ -12,6 +12,8 @@ export const HTML_LANGUAGES = [
   'hbs',
   'html',
   'HTML (Eex)',
+  'HTML (EEx)',
+  'html-eex',
   'jade',
   'leaf',
   'liquid',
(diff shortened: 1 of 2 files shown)


zolrath:
Ooh I actually see that the uppercase EEx was added to `master` but isnt in `next`.
Maybe thats why v0.3.0-alpha.1 didn't work with Elixir?

bradlc:
Hey @zolrath. I noticed that `HTML (EEx)` was missing but didn't know about `html-eex`, so thank you!

zolrath:
`html-eex` isn't officially merged into the project at the moment but it should be the id used to become compliant with the Language ID guidelines so it should be handy in the near future!

PR #116, opened 2020-05-10
Add SugarSS support

danmichaelo:
[SugarSS](https://github.com/postcss/sugarss) is an indent-based CSS syntax for PostCSS.

SugarSS MIME-type is `text/x-sugarss` with `.sss` file extension.


bradlc:
Thanks!

PR #122, opened 2020-05-25
Adds [%tw syntax for use with tailwind-ppx

celsobonutti:
Adds the [%tw syntax for use with [tailwind-ppx](https://github.com/dylanirlbeck/tailwind-ppx).

Although I added the regex (and it's apparently right, based on regex tests) I'm not able to test the extension locally, as it seems that my changes don't have any effect.

I tried using `npm run package` and installing the VSIX file, does anyone have any idea on what could be happening or can build the extension and test it?

### DIFF
--- a/src/lsp/util/getClassNameAtPosition.ts
+++ b/src/lsp/util/getClassNameAtPosition.ts
@@ -12,7 +12,7 @@ export function getClassNameAtPosition(
   }
   const text1: string = document.getText(range1)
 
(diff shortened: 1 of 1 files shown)


dylanirlbeck:
@bradlc This PR has been open for a while, and it's only two lines. Would you mind taking a look? Thanks so much!!

dylanirlbeck:
For posterity's sake, the improvement that this PR would make will be subsumed by the feature proposed in #129.

Issue #129, opened 2020-06-16
Custom class name completion contexts


bradlc:
There has been quite a few requests for the extension to support class name completions in contexts other than a standard class(Name) attribute.

Some examples:

- [`tailwind.macro`](https://github.com/bradlc/babel-plugin-tailwind-components) and [`twin.macro`](https://github.com/ben-rogerson/twin.macro) (#46, #123):
  ```
  tw`_`
  tw.div`_`
  ```
- [`tailwind-rn`](https://github.com/vadimdemedes/tailwind-rn) (#99):
  ```
  tailwind('_')
  ```
- [`tailwind-ppx`](https://github.com/dylanirlbeck/tailwind-ppx) (#120):
  ```
  <Component className=[%tw "_"] />
  ```

I just wanted to consolidate all of these requests into a single issue that can be tracked more easily, as I think the solution could be the same for each of them.

I am reluctant to hard-code each of these cases into the extension because none of them are "official" methods of using Tailwind, and it may become a maintenance burden.

However, I am open to the idea of adding a user setting which would allow the definition of custom regular expressions. For example for `tailwind-rn` your regular expression might be something like: `/\btailwind\([^)]+/ig`

If you're interested in this feature feel free to "watch" this issue for updates, and post any comments/suggestions you may have.

FredyC:
The `twin.macro` also supports `<div tw="text-black" />`, can you add it to the list, please?

tobkle:
Thank you @bradlc. Would it make sense to have those regex's as an array in the vscode settings? It'll be more extensible though.

bradlc:
> The `twin.macro` also supports `<div tw="text-black" />`, can you add it to the list, please?

Added!

> Thank you @bradlc. Would it make sense to have those regex's as an array in the vscode settings? It'll be more extensible though. I think it's that what you meant with user settings. That's a good idea.

Yes exactly. It would be a user setting.

johot:
In my case I pull some tailwind class combinations into my own config objects. Could one solution be to simply add a comment above these variables and make it work?

For example a `// @tw` comment like below:

```
const theme: Theme = {
    borders: borders,
    input: {
      // @tw
      default:
        " border rounded-lg bg-white shadow-light-md  " + borders.default,
    }
}
```

maykefreitas:
[...]

PR #155, opened 2020-07-21
Added support for fable/fsharp

### LINKED ISSUE #154: Support for Fable & Feliz
Hi there,

I am trying to set the plugin up for [Fable](https://fable.io/) and [Feliz](https://github.com/Zaid-Ajaj/Feliz).

The setting seems to work:

```json
  "fsharp": "html"
}
```

image

However, instead of having the completion "after the dot", I would need it to work inside of `prop.className ""`.

Is there a way to make this work?

CallumVass:
Should fix #154

### DIFF
--- a/src/lib/languages.ts
+++ b/src/lib/languages.ts
@@ -6,6 +6,7 @@ export const DEFAULT_LANGUAGES = [
   'edge',
   'ejs',
   'erb',
+  'fsharp',
   'gohtml',
   'GoHTML',
   'haml',
--- a/src/lsp/util/find.ts
+++ b/src/lsp/util/find.ts
@@ -129,7 +129,7 @@ export function findClassListsInHtmlRange(
   range?: Range
 ): DocumentClassList[] {
   const text = doc.getText(range)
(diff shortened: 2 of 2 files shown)


Zaid-Ajaj:
@bradlc Can you please have a look at the PR? 🙏 This small improvement would make a big difference for F# people using tailwind. In the meantime, maybe @CallumVass can resolve the merge conflicts?

CallumVass:
Sure - I'll take a look although there weren't any merge conflicts when I submitted this 4 MONTHS AGO with zero acknowledgement or comment.

bradlc:
Hey, thanks for taking the time. I think I would prefer for this to be handled by #129. We can't support absolutely every use-case out-of-the-box and we have to draw the line somewhere. I think that line is roughly "things that look like standard HTML attributes."

Check out [my comment](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/154#issuecomment-740608084) on #154 for details of how the new `tailwindCSS.experimental.classRegex` setting can be used to support Feliz syntax, thanks!

Issue #172, opened 2020-08-20
Intellisense not working in angular project


alexbjorlig:
Trying to setup tailwind with our Angular project, and it does not work 🤔

**tailwind.config.js**
```js
module.exports = { theme: { extend: {} }, variants: {}, plugins: [] };
```

**webpack.config.js**
```js
module.exports = {
  module: {
    rules: [
      {
        test: /\.scss$/,
        loader: 'postcss-loader',
        options: {
          ident: 'postcss',
          syntax: 'postcss-scss',
          plugins: () => [
            require('postcss-import'),
            require('tailwindcss')('./tailwind.config.js'),
            require('autoprefixer'),
          ]
        }
      }
    ]
  }
};
```

**package.json**
```
"devDependencies": {
        "tailwindcss": "^1.6.2"
    },
```

I followed the troubleshooting steps from the readme, but no help. Is there somewhere we can trace logs/issues?

bradlc:
Hey @alexbjorlig. Do you see any errors in the `Output` panel? You can open this from the command palette or press `cmd+shift+u` if you're on macOS. (You'll need to select `tailwindcss-intellisense` from the dropdown)

Also are you able to share your project at all? If not could you provide some details such as package manager (npm, yarn etc.), operating system, VS Code version, general project structure?

alexbjorlig:
I don't see the extension at all in the `Output` panel?

<img width="1466" alt="Screenshot 2020-08-25 at 20 08 36" src="https://user-images.githubusercontent.com/9967422/91211282-d0fcac80-e70e-11ea-8f47-f79832061beb.png">

And it is installed? 

<img width="1328" alt="Screenshot 2020-08-25 at 20 09 27" src="https://user-images.githubusercontent.com/9967422/91211331-e7a30380-e70e-11ea-8a14-d10451c1d813.png">

Some details:

npm: 6.14.7
node: v14.8.0
OS: macOS Catalina 10.15.6 (19G73)
VS Code Version: 1.48.2
General project structure: Angular CLI project

alexbjorlig:
I just tried to verify if the extension would work in a "clean" Angular CLI environment - and indeed it does. How can I proceed from here? If I buy you a beer - would you mind take a quick look at our project, or do you know if I can get some error output somehow?

bradlc:
[...]

PR #179, opened 2020-08-26
Added section to troubleshooting

alexbjorlig:
In refrence to issue #172, this VS code setting:
```
"search.exclude": {
    "**/*.js": true
},
```
prevented the Tailwind config from being found ☺️


bradlc:
Thanks @alexbjorlig!

Issue #118, opened 2020-05-16
IntelliSense not works for Typescript


gkn06:
How to make it work for .ts files?

I think `typescript `should be added here
https://github.com/bradlc/vscode-tailwindcss/blob/f5dfe02f74ac9bd68529f1997ae875691b819833/src/index.ts#L34-L39

gkn06:
tried alpha version too with this setting but still no luck

```json
 "tailwindCSS.includeLanguages": {
    "plaintext": "typescript"
  }
```

bradlc:
Thanks, I'll add it!

Your `includeLanguages` setting should look like this by the way:

```
"tailwindCSS.includeLanguages": {
  "typescript": "javascript"
}
```

gkn06:
Thanks @bradlc

tobkle:
For Typescript React I need this in the settings:
```javascript
"tailwindCSS.includeLanguages": {
    "typescript": "javascript",
    "typescriptreact": "javascript"
}
```

Issue #72, opened 2019-09-23
Doesn't work for TSX files


arimendelow:
When using TypeScript with React, this plugin doesn't work. How can I configure it to look at .tsx files?

lednhatkhanh:
How did you solve this?

or2008:
just installed and it doesn't work for me, lets see if this comment fixes it :)

Manubi:
doesn't work for me either in .tsx only if I write "class" instead of "className" I get the autosuggest...

arimendelow:
That's weird, it should work without any configuration. I'm still unsure why it didn't for me initially.

sw-yx:
same - it didnt work on tsx originally, then i panicked and clicked around and now it works

bentruyman:
I also ran into this. Not sure if it's anything TSX-related (but it was failing in TSX files for me). But to fix it I simply created a `tailwind.config.js` file in my project.

```js
module.exports = {
  theme: {},
  variants: {},
  plugins: [],
};
```

It seems [a config file](https://github.com/bradlc/vscode-tailwindcss/blob/f5dfe02f74ac9bd68529f1997ae875691b819833/src/index.ts#L24) is what enables this extension in a project.

exoer:
In vscode - extensions - Tailwin CSS Intellisense. Click on the settings button ("manage"). 

Find Tailwind CSS - Include languages 
Click 'edit in settings.json'

add 
`
  "tailwindCSS.includeLanguages": {      
    "plaintext": "javascript"
  }
`

b-groeneveld:
I've added 

```"tailwindCSS.includeLanguages": { "typescriptreact": "html" },```

And restarted vscode and it was good to go

oxyian-dev:
> I've added
> 
> `"tailwindCSS.includeLanguages": { "typescriptreact": "html" },`
> 
> And restarted vscode and it was good to go

Super its working for me

khalidahmada:
You can add this in user settings

`"tailwindCSS.includeLanguages": {
    "typescript": "javascript",
    "typescriptreact": "javascript"
}`

miko1991:
doesnt work with custom webpack babel typescript react

avxkim:
@khalidahmada this what it made work for me in Nuxt 3 + TS

paulwongx:
> You can add this in user settings
> 
[...]

PR #189, opened 2020-09-13
Add example of how to support typescript, and typescript + react

cktang88:
This seems to be a common problem, and the docs could help make this clearer (eg. #118 #72 )


bradlc:
Hey @cktang88. `typescript` and `typescriptreact` are supported out-of-the-box so this is not required. Changing this setting causes the extension to "reboot," which is why is can give the illusion of fixing an issue when really it is unrelated.

PR #196, opened 2020-09-22
Add hyperscript support, enable 'key: value' style js code

ntilwalli:
This is a simple regex update to enable autocomplete support for hyperscript.  Previously only `class="foo"` or `className="foo"`, would activate the autocomplete, with this PR `class:"foo"`, `class: "foo"`, `className:"foo"` and `className: "foo"` also activate autocomplete.

### DIFF
--- a/src/lsp/providers/completionProvider.ts
+++ b/src/lsp/providers/completionProvider.ts
@@ -130,7 +130,7 @@ function provideClassAttributeCompletions(
     end: position,
   })
 
-  const match = findLast(/(?:\b|:)class(?:Name)?=['"`{]/gi, str)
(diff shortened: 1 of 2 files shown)


bradlc:
Hey @ntilwalli. Thank you, but I think for now I'd like to keep the built-in support to things that look like HTML class attributes (i.e. they use `=` to separate the key and value).

You could perhaps enable support for this syntax by using the [`experimental.classRegex` setting](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/129#issuecomment-735915659). Here's a basic example:

```
"tailwindCSS.experimental.classRegex": [
  "\\sclass:\\s*\"([^\"]*)\""
]
```

PR #211, opened 2020-11-13
Change class regex to match elm source

kdmurthy:
Just a small modification to the regex for use with elm source `class` method.

### DIFF
--- a/src/lib/languages.ts
+++ b/src/lib/languages.ts
@@ -5,6 +5,7 @@ export const DEFAULT_LANGUAGES = [
   'django-html',
   'edge',
   'ejs',
+  'elm',
   'erb',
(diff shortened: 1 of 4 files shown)


bradlc:
Hey @kdmurthy. Thank you, but I think for now I'd like to keep the built-in support to things that look like HTML class attributes (i.e. they use `=` to separate the key and value).

You could enable support for elm using a combination of the [`includeLanguages`](https://github.com/tailwindlabs/tailwindcss-intellisense#tailwindcssincludelanguages) and [`experimental.classRegex`](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/129#issuecomment-735915659) settings

PR #237, opened 2020-12-31
Adding Tailwind Custom Data Extension for CSS

samtrion:
With this PR the tailwind specific @-rule are declared to VSCode as explained here - [Custom Data Extension](https://code.visualstudio.com/api/extension-guides/custom-data-extension). So that they are no longer marked as incorrect by the VSCode CSS Language Service.

Further information about Custom Data Extension
[Description of the Format](https://github.com/microsoft/vscode-css-languageservice/blob/master/docs/customData.md)
[Microsoft Example for Implementation](https://github.com/microsoft/vscode-extension-samples/tree/master/custom-data-sample)

### DIFF
--- a/packages/tailwindcss-intellisense/package.json
+++ b/packages/tailwindcss-intellisense/package.json
@@ -167,6 +167,11 @@
           "type": "boolean"
         }
       }
+    },
+    "css": {
+      "customData": [
+        "./tailwind.css-data.json"
+      ]
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


bradlc:
Wow I had no idea you could do this! Thanks for the PR @samtrion 

I think on this occasion I'm going to say no for a couple of reasons:

- Although this suppresses "unknown at rule" warnings, it does not suppress `@apply` class list errors ("semi-colon expected", "at-rule or selector expected")
- Adding the at directives using custom data provides suggestions, but these are less flexible than the existing suggestions the extension provides. For example we cannot provide different suggestions based on the user's Tailwind version. The existing suggestions provide links to the correct docs site based on the Tailwind version you are using.

Thanks again!

PR #261, opened 2021-02-08
add '-' to triggerCharacters

merrickluo:
Hi, thx for the great project! I hope we can add `-` to the completion trigger characters, it seems not a problem in vscode, but when I use it in [emacs with lsp-mode](https://github.com/merrickluo/lsp-tailwindcss), it can't trigger completion when type `-`, seems like it's been treated as a word separator. I can workaround by adding that in elisp code, but I'm hoping we can just add it to the server if there is no big trouble.

### DIFF
--- a/packages/tailwindcss-language-server/src/server.ts
+++ b/packages/tailwindcss-language-server/src/server.ts
@@ -92,6 +92,8 @@ const TRIGGER_CHARACTERS = [
   '!',
   // JIT opacity modifiers
   '/',
+  // class name including '-'
+  '-',
 ] as const
 
 const colorNames = Object.keys(namedColors)


packages/tailwindcss-intellisense/src/lsp/server.ts
```diff
@@ -156,6 +156,8 @@ connection.onInitialize(
             '.',
             // config/theme helper
             '[',
+            // class name including '-'
+            '-',
             // TODO: restart server if separater changes?
```
gekkedev, line 161:
```suggestion
            // TODO: restart server if separator changes?
```
since you already touch the preceding line, I believe it's a good opportunity to fix the typo :wink:

gekkedev (changes requested):

packages/tailwindcss-intellisense/src/lsp/server.ts
merrickluo: sure...but the whole file seems gone after I did a rebase, updated PR in the new file 🙂️.

gekkedev: Alright then nevermind

gekkedev (approved):

PR #262, opened 2021-02-09
Make output chanel name more human-friendly

neupauer:
According to the documentation, the channel name is: "Human-readable string which will be used to represent the channel in the UI". So it shouldn't break anything.

Change: `tailwindcss-intellisense` -> `Tailwind CSS IntelliSense`

Screen Shot 2021-02-09 at 14 17 11

Links: 
[VS Code API Documentation](https://code.visualstudio.com/api/references/vscode-api#window) / FUNCTIONS / createOutputChannel


bradlc:
Thank you!

Issue #91, opened 2019-12-30
tailwindcss-language-server debug log conflicting with vim-lsc


mikepjb:
Hi Brad,

I've been using the underlying language server for this plugin using vim-lsc (see raised issue here: https://github.com/natebosch/vim-lsc/pull/251)

There is a call to `console.log(t.configDependencies)` that is causing a `[]` to be prepended to the message header being sent. This prevents `tailwindcss-language-server` being used.

It'd be good to understand why that log is there, if it's not useful could you remove it from the npm library?

Best,
Mike

mikepjb:
Bumping this, any chance we can sort something out?

Let me know if there's anything I can do to help :+1:

bradlc:
All logs have been removed in [`v0.3.0`](https://github.com/bradlc/vscode-tailwindcss/releases/tag/v0.3.0), which is available now

PR #263, opened 2021-02-11
Remove logging conflicting with language server output

ElArtista:
Similar to issue #91 opened and solved in the past.
We must not put console.log stdout outputs as this conflicts with the language server when running on --stdio mode.

Proposed fix tested and working with vim-lsp.


bradlc:
Thanks @ElArtista. I didn't want to remove these logs so [I have overridden `console.log` and `console.error`](https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/tailwindcss-intellisense/src/lsp/server.ts#L46-L47), based on code from [vscode-json-languageserver](https://github.com/vscode-langservers/vscode-json-languageserver/blob/master/src/node/jsonServerMain.ts#L18-L19). This seems to fix the issue in Nova but let me know if you're still running into issues.

Issue #274, opened 2021-03-09
Build is broken


ElArtista:
First of all providing some build instructions would be really appreciated.

Tried to clone and build the `tailwindcss-intellisense` module with:
```
npm install
cd packages/tailwindcss-intellisense/
npm run build
```

And I'm getting:
```
Error: [tsl] ERROR in /code/tailwindcss-intellisense/packages/tailwindcss-intellisense/src/lsp/server.ts(57,3)
      TS2741: Property 'showPixelValues' is missing in type '{ classRegex: undefined[]; }' but required in type '{ classRegex: string[]; showPixelValues: boolean; }'.
```

Seems that its trying to build using tailwindcss-language-service 0.0.12 and not the local version of the package

bradlc:
Closing this because it is outdated now. Build should work with `npm run build` 👍

PR #287, opened 2021-03-29
Fix build due to invalid default `Settings`

misha:
As mentioned in #274, a file was preventing an errror-less build because an experimental setting was missing. I supplied a default value for the field. 

However, upon fixing it I found that the underlying type `Settings` had been compiled with some previous version of the type that failed to includes the following properties:

1. `showPixelEquivalents`
2. `rootFontSize`

The file that declares the type, `state.ts`, was already updated to hold those properties, so I re-built the language server. This fixed the two fields above, but the experimental setting was gone this time. To fix the final error, I added that field to the original type definition in `state.ts`.

Likely to completely fix this, you would need to re-release the language server and update the extension's version.

### DIFF
--- a/packages/tailwindcss-intellisense/src/lsp/server.ts
+++ b/packages/tailwindcss-intellisense/src/lsp/server.ts
@@ -56,6 +56,7 @@ const defaultSettings: Settings = {
   includeLanguages: {},
   experimental: {
     classRegex: [],
+    showPixelValues: false,
   },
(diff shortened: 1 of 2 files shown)


ElArtista:
Bumped into this too, waiting a long time to get this fixed

bradlc:
Hey @misha. Sorry for the delay. I don't think this is relevant anymore, but thanks anyway! 🙏

PR #317, opened 2021-05-10
Restore workspaceContains activation

polyzen:
For compatibility with clients other than VS Code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
   },
   "icon": "media/icon.png",
   "activationEvents": [
-    "onStartupFinished"
+    "onStartupFinished",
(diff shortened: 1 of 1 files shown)


polyzen:
This was not enough -- must have had [a copy of `0.5.10` installed](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tailwindcss-intellisense) when I attempted to try this.

For reference, I use vim-lsp:
https://github.com/mattn/vim-lsp-settings/blob/b37de024025cc34ee5f422fcf3638da99851e243/settings/tailwindcss-intellisense.vim

PR #341, opened 2021-05-30
fix crash on not passing "userLanguages" option

rchl:
Since the server can be used by various clients and there is no
guarantee that the "userLanguages" are provided, handle that
case gracefully and don't crash.


bradlc:
Thanks! 🙏

PR #351, opened 2021-06-17
Suggest lighter css rule

sachinraja:
`css.validate` disables all css linting. This rule only disables linting for the unknown `@` rules like `@tailwind`.

### DIFF
--- a/packages/vscode-tailwindcss/README.md
+++ b/packages/vscode-tailwindcss/README.md
@@ -34,10 +34,10 @@ Provides syntax definitions so that Tailwind features are highlighted correctly.
 
 ## Recommended VS Code Settings
 
-VS Code has built-in CSS validation which may display errors when using Tailwind-specific syntax, such as `@apply`. You can disable this with the `css.validate` setting:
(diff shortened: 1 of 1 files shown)


bradlc:
Interesting! I swear at some point there was a built-in rule that caused an error when you had something like this:

```css
.example {
  @apply one two;
}
```

You would get an error like "expected semi-colon after `one`", and the only way to disable that one was to disable `css.validate` completely. But it seems like this error doesn't happen anymore (or I imagined it), so I _think_ this change is good 🤔

sachinraja:
I believe I was getting the same error, but it just mysteriously disappeared. The only thing I can find is [this PR](https://github.com/microsoft/vscode-css-languageservice/pull/10), but it's from 2016.

bradlc:
Hey @sachinraja. For now we're going to leave it as it is, because there's other situations where you get lint errors if `css.validate` is `true`, for example when using the [`screen` function](https://tailwindcss.com/docs/functions-and-directives#screen):

```css
@media screen(sm) {}
```

Thank you for the suggestion though!

PR #365, opened 2021-07-06
Avoid `invalidConfigPath` for function values

choplin:
Pass `validateConfigPath` if a type of value equals to `function`. The purpose of this PR is to avoid unwanted warnings when we define theme values as functions in tailwind.config.js, which, for example, are required to define theme colors with opacity enabled. https://github.com/tailwindlabs/tailwindcss/discussions/2206#discussioncomment-60081


bradlc:
Nice catch, thank you!

Issue #456, opened 2022-01-07
IntelliSense not working with Styled Components


hugohutri:
Tailwind CSS IntelliSense version:  v0.7.4
Tailwind CSS version: v3.0.11
Package manager: yarn
Operating system: Ubuntu

Intellisense does not work with [styled-components](https://styled-components.com/).
![image](https://user-images.githubusercontent.com/55588133/148538332-871b2abb-32af-43dd-9ab2-f3c480aa7af6.png)

The RegEx doesn't allow the use of `:`-symbol after className. E.g: `className: "abc"`.
If the `:` is replaced with `=`, the suggestions will be visible.

![image](https://user-images.githubusercontent.com/55588133/148538501-a83666d7-1698-4368-93dd-247ee8fb1217.png)

### Solution
Add `:` to the RegEx

bradlc:
Hey @hugohutri. Please take a look at my comment on #457 for my thoughts on this one 👍

PR #457, opened 2022-01-07
fix/regex-with-colon

hugohutri:
Allow IntelliSense to suggest classes with Styled Components by allowing the use of ":"-symbol. e.g. className: "ABC"

Fixes the issue #456


bradlc:
Hey @hugohutri. Thank you, but I think for now I'd like to keep the built-in support to things that look like HTML class attributes (i.e. they use `=` to separate the key and value).

You could use the [`experimental.classRegex`](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/129#issuecomment-735915659) setting to enable support for other use-cases. Check out my comment here for more information: https://github.com/tailwindlabs/tailwindcss-intellisense/issues/129#issuecomment-735915659

PR #461, opened 2022-01-08
Prevent errors in not semicolon languages

### LINKED ISSUE #460: Syntax highlight problem in SASS format

For example: v0.7.5

For example: v3.0.12

For example: yarn

For example: Manjaro

**Reproduction URL**

https://github.com/lucasvazq/tailwindcss-intellisense-error

**Describe your issue**

I was working on a new project, to which I added tailwindcss and decided to install this extension for VSCode, and I get an undesired error.

<blockquote>

Screenshot_2022-01-08_23-00-55
```
'base
@tailwind components
@tailwind utilities

@import url("https://fonts.googleapis.com/css2?family=Poppins:wght@400&display=swap")
' is not a valid value. Did you mean 'base'?(invalidTailwindDirective)
```
</blockquote>

The first thing I did was try to fix it with the editor `Quick Fix -> Replace with 'base'`. The result was the following:

<blockquote>

Screenshot_2022-01-08_23-11-30
</blockquote>

After that, I wanted to add a new line and started seeing the problem again:

<blockquote>

Screenshot_2022-01-08_23-04-32
```
'base
' is not a valid value. Did you mean 'base'?(invalidTailwindDirective)
```
</blockquote>

So, I started testing, and I realized that the problem is because the extension expects that always after the tailwind directives there be a semicolon.

<blockquote>

Screenshot_2022-01-09_01-49-32
</blockquote>

This is okay for most syntaxes, but I'm using SASS (not SCSS), and I don't need to add a semicolon.

Related code: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/tailwindcss-language-service/src/diagnostics/getInvalidTailwindDirectiveDiagnostics.ts#L34

lucasvazq:
## Related

Resolves #460

## Comments

Prevent unwanted error messages related to the tailwind directive in languages that don't require semicolons at the end of lines.

Support added for:
- sass
- sugarss
- stylus

List taken from: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/master/packages/tailwindcss-language-service/src/util/languages.ts#L34

### DIFF
--- a/packages/tailwindcss-language-service/src/diagnostics/getInvalidTailwindDirectiveDiagnostics.ts
+++ b/packages/tailwindcss-language-service/src/diagnostics/getInvalidTailwindDirectiveDiagnostics.ts
@@ -27,7 +27,7 @@ export function getInvalidTailwindDirectiveDiagnostics(
     ranges.push(...boundaries.css)
   }
 
-  const notSemicolonLanguages = ["sass"]
+  const notSemicolonLanguages = ["sass", "sugarss", "stylus"]
   let regex: RegExp
   if (
     notSemicolonLanguages.includes(doc.languageId) ||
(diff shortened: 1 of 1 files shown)


bradlc:
This is great, thank you! I just made a couple of minor changes 👍

PR #467, opened 2022-01-16
Support completion in template strings using `

ModProg:
closes #467


bradlc:
Please see https://github.com/tailwindlabs/tailwindcss-intellisense/issues/465#issuecomment-1016705998

ModProg:
For anyone coming across this needing this as regex you can use:
```
'(?:classes|class)\\s*=\\s*`([^`]*)',
```

Replace `classes|class` with all the variable names you want to use.

Still would have considered this a small enough change to be added in the code IMO.

PR #472, opened 2022-01-31
Consider `.ts` file extensions on enhanced resolver

JensDll:
Hey, I've been using TypeScript to write my `tailwind.config.ts` file and ran into an issue when resolving imports. Using imports like `import { foo } from './foo'` where `foo` is some TypeScript file will cause an exception. Thrown here:

https://github.com/tailwindlabs/tailwindcss-intellisense/blob/85cf5edccbcf801052d3bedf68e09165796993de/packages/tailwindcss-language-server/src/util/resolveFrom.ts#L22

tailwind-extension-error

I have a small reproduction to play around with, which uses [esbuild-register](https://github.com/egoist/esbuild-register) to load the config file:
https://github.com/JensDll/tailwindcss-config-reproduction

The problem seems to be that [enhanced-resolve](https://github.com/webpack/enhanced-resolve) does not consider `.ts` file extensions per default. Additionally, TypeScript does not allow you to specify them explicitly at the moment. Related discussion: 
- https://github.com/microsoft/TypeScript/issues/37582

This pull request will add `.ts` to the enhanced resolver's `extensions` option. The reproduction should no longer throw an error.

### DIFF
--- a/packages/tailwindcss-language-server/src/util/resolveFrom.ts
+++ b/packages/tailwindcss-language-server/src/util/resolveFrom.ts
@@ -5,9 +5,9 @@ function createResolver(options: Partial<ResolveOptions> = {}): Resolver {
   return ResolverFactory.createResolver({
     fileSystem: new CachedInputFileSystem(fs, 4000),
     useSyncFileSystemCalls: true,
-    // cachePredicate: () => false,
     conditionNames: ['node', 'require'],
-    ...options,
+    extensions: ['.js', '.ts', '.json', '.node'],
+    ...options
   })
 }
 


bradlc:
Hey @JensDll. I'm doing a bit of issue/PR cleanup, and I am reluctant to merge this one, because Tailwind CSS (and by extension, Tailwind CSS IntelliSense) doesn't support TypeScript config files. I totally understand that you're working around this by using a _JavaScript_ config file in conjunction with `esbuild-register`, but I think supporting this quite specific setup is the wrong approach for us right now. I wouldn't want to add support for this, and then have to remove it when/if we add proper TypeScript config support to Tailwind CSS (this is on our radar by the way!). I hope that's understandable, and thank you for the suggestion! 🙏

PR #521, opened 2022-04-13
Convert dlv require statements to imports

remcohaszing:
It was already imported in one place, that appears to work fine.

Using `require` results in invalid ESM output, because `require` is not defined in ES modules.


bradlc:
Thank you!

Issue #407, opened 2021-09-16
Add Elixir Phoenix 1.6 .html.heex templates support


malloryerik:
Hi, the newest version of Phoenix uses a new template called [.html.heex](https://hexdocs.pm/phoenix_live_view/assigns-eex.html). 

What can I do to add this to the extension (which is great, btw)?
Is it just a matter of adding a line to [`packages/tailwindcss-intellisense/src/lib/languages.ts`](https://github.com/tailwindlabs/tailwindcss-intellisense/blob/8fce24db9c480a9615e175fb3acaac91f7a0a7c1/packages/tailwindcss-intellisense/src/lib/languages.ts)?

Thanks.

bradlc:
Hey @malloryerik. Yes exactly, we just need to know the "language ID" for that file type. You can typically find this by looking at the `Feature Contributions` panel of the extension that provides the language support. I cannot find an extension that adds support for `.html.heex` files though. Here's what `vscode-elixir` contributes:

![image](https://user-images.githubusercontent.com/2615508/133767115-b1d7f738-3ac1-49db-bbb4-47beb342b06a.png)

Are you aware of an extension that provides support for `.html.heex` files?

malloryerik:
Oh I see, thanks very much.

I believe the language id is `phoenix-heex`.

The Elixir and Phoenix core teams have a new extension, [Phoenix Framework](https://github.com/phoenixframework/vscode-phoenix)


<img width="703" alt="Screen Shot 2021-09-18 at 11 34 07 AM" src="https://user-images.githubusercontent.com/10395849/133869470-ecaef498-2b02-452a-81f8-a71539bd5c10.png">

Btw the most popular (and useful) Elixir / Phoenix plugin is [ElixirLS](https://github.com/elixir-lsp/vscode-elixir-ls). They've also [added `phoenix-heex`](https://github.com/elixir-lsp/vscode-elixir-ls/pull/204/commits/a079091f69c383472764b743cf017dd1c1a728a7).

bradlc:
Thanks! `phoenix-heex` has been added in `v0.6.15` 👍

PR #597, opened 2022-08-27
Support Surface templates

joshprice:
Surface support is defined here https://github.com/msaraiva/vscode-surface/blob/master/syntaxes/surface.json

See https://github.com/tailwindlabs/tailwindcss-intellisense/issues/407 and https://github.com/tailwindlabs/tailwindcss-intellisense/commit/1253902e88d085ee82e0e4abb4e60de17fa925ab


bradlc:
Thanks!

PR #602, opened 2022-09-12
fix: guard against optional client capabilities

merrickluo:
for lsp client like lsp-mode for emacs which doesn't have colorProvider, server is failing to start, and according to the specification, the capabilities here is all optional, so we should guard the nil situation.


takueof:
Hello @bradlc and other contributors, I am one of the people who is very troubled by this problem.
See https://github.com/merrickluo/lsp-tailwindcss/issues/44 for previous discussion.
Thank you.

bradlc:
Thank you!

PR #635, opened 2022-10-19
Handle the `DEFAULT` case for `matchVariant`

RobinMalfait:
This PR handles the `DEFAULT` value for `matchVariant`.

It should show you `foo:` instead of `foo-DEFAULT:` in completions.

---

@bradlc, I'm not 100% sure how to test this so going to leave this open until whenever you have time
to look at it.


bradlc:
Thank you!

PR #645, opened 2022-10-28
Remove `typescript` from list of languages

### LINKED ISSUE #644: Intellisense should not work in `.ts` files

v0.9.1

v3.2.0

pnpm

Windows

**Tailwind config**

```js
module.exports = {
  content: ["./src/**/*.{js,ts,jsx,tsx}"],
  theme: {
    extend: {},
  },
  plugins: [],
};
```

**VS Code settings**

```json
// So many settings. Only this setting enabled from extension
{
  "tailwindCSS.emmetCompletions": true,
}
```

**Describe your issue**

HTML or JSX isn't written in `.ts` files. So I think the extension should not show the intellisense in the `.ts` files.
Intellisense works in `.tsx` files and that's good but it shouldn't in `.ts` files

image

max-programming:
I hope this fixes #644

### DIFF
--- a/packages/tailwindcss-language-service/src/util/languages.ts
+++ b/packages/tailwindcss-language-service/src/util/languages.ts
@@ -49,7 +49,6 @@ export const jsLanguages = [
   'javascriptreact',
   'reason',
   'rescript',
-  // 'typescript',
   'typescriptreact',
 ]
 
--- a/packages/tailwindcss-language-service/src/util/languages.ts
+++ b/packages/tailwindcss-language-service/src/util/languages.ts
@@ -49,7 +49,7 @@ export const jsLanguages = [
   'javascriptreact',
   'reason',
   'rescript',
-  'typescript',
+  // 'typescript',
   'typescriptreact',
 ]
 


max-programming:
@bradlc What do you think about this?

bradlc:
I agree that it makes sense to remove this, thanks!

SEQUOiA87:
Ehhh 👀

This removes typescript from the `jsLanguages` constant. Are you sure about this?

I don't specifically know how this constant is used by the plugin, but the fact that:
- typescript is of course a "javascript language" (at least not less than those other languages which are still part of the constant: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/v0.9.2/packages/tailwindcss-language-service/src/util/languages.ts#L49-L52)
- in Angular, it's perfectly fine to define an Angular component's template (HTML) code directly within the component, i.e. within the .ts file (see here: https://angular.io/guide/component-overview#defining-a-components-template)

... let's me think that this change should be reverted.

In the meantime, for anyone needing this plugin to also work in typescript files, putting this into the VS Code settings helps:
```
    "typescript": "javascript"
}
```

max-programming:
Oh I see that now. Sorry @SEQUOiA87 I totally forgot Angular in this case.

Maybe something can be done to detect if there's an Angular project and enable then push `typescript` in the array?

max-programming:
Or I guess in the meantime, we could enable `typescript` while an issue like this persists?

bradlc:
Yeah I think I'll revert this for now. Was already considering it since seeing #687 👍

PR #656, opened 2022-11-09
Fixed error when initializationOptions is undefined

anbcodes:
This PR fixes an issue where the language server crashes if the `initalizationOptions` property is not defined. The `initializationOptions` property is optional according to the [language server protocol specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize), so the server shouldn't rely on it being an object.

This also fixes https://github.com/helix-editor/helix/issues/2213. Helix does not send the `initializationOptions` by default and so the language server was crashing.


bradlc:
Thank you!

PR #658, opened 2022-11-10
feature: hide/show class attributes

### LINKED ISSUE #313: Feature request: toggle show/hide classnames
Hey,
Tailwind is awesome. A minor downside though is that my HTML is much denser, less easy to scan, because of all the new class names.

I'd like a way to collapse all Tailwind classes, while keeping visible custom classes, such that it becomes easier to scan when I need to focus on my HTML. A keyboard shortcut to toggle visibility would be great (I think), e.g.

turn this:

```html
<div class="flex items-start self-center mt-3">
  <div class="flex items-center h-5">
    <input name="yo" type="checkbox" class="w-4 h-4 text-red-600 border-gray-300 rounded focus:ring-indigo-500 my-custom-class" value="1">
  </div>
</div>
```

to this:

```html
<div class="•••">
  <div class="•••">
    <input name="yo" type="checkbox" class="••• my-custom-class" value="1">
  </div>
</div>
```

What do you think?

mtbaqer:
### This is a Draft PR, the purpose of it is to work as a starting point, the code here is nowhere near being production-ready

Given the progress that was made in the area by [vscode-inline-fold](https://github.com/moalamri/vscode-inline-fold) and [vscode-explicit-folding](https://github.com/zokugun/vscode-explicit-folding), I wanted to make this Draft PR to provide an approach of implementing hiding and showing class attributes in the tailwind extension. This hopefully works as a starting point for the feature as well as to reopen the discussion on the topic.

Hoping to reopen issue #313 to work as a place for the conceptual discussion while keeping it here only for implementation details.

Disclaimer: Many elements of this PR were taken from the extensions above with heavy modifications to make them work with the rest of the extension. 

## Issues

- Fixes #313 
- Could potentially fix https://github.com/tailwindlabs/tailwindcss/issues/6897 with minor tweaks

## Demo

https://user-images.githubusercontent.com/32401657/200982868-58bdbd96-33cc-4be1-8766-49c1416b1774.mov

## Features

- [x] Inline fold of class attributes (design inspired by https://github.com/tailwindlabs/tailwindcss/discussions/7922)

Screenshot 2022-11-09 at 20 40 26

Screenshot 2022-11-09 at 20 40 31

- [x] Multiline fold using VSCode native folding.

Screenshot 2022-11-09 at 20 57 04

Screenshot 2022-11-09 at 20 57 18

- [x] Toggle command (allowing for a keyboard shortcut) to enable/disable hiding class attributes.
- [x] Auto show class content when cursor is within range or on selection.
- [x] Multi-cursor support.
- [x] Support current `tailwindCSS.classAttributes` extension setting.
- [x] New `tailwindCSS.experimental.foldClassAttributes` settings value to disable the feature.

## TODO

- [ ] Support `tailwindCSS.includeLanguages` setting.
- [ ] Allow an option to disable the auto unfolding, to make classes stay folded until toggled back.
- [ ] Fix known bugs:
  - [ ] When user manually unfolds a multiline class attribute, the first line stays inline folded
  - [ ] When user disables folding, multiline class attributes folded in documents beside the current one become "normally" folded by VSCode instead of unfolding.
- [ ] Reorganize current code (didn't know where to put the logic, so put everything in one file and added the calls at the very bottom of `extension.ts`).


kossnocorp:
I just started working on a similar thing. I'm glad @adamwathan pointed me to it. It's such a fantastic idea, I can't wait for you guys to build it!

surjithctly:
@mtbaqer Possible to make sure to see the `=""` part all the time? Currently, it feels weird. So it would look like this:

```html
<!-- ⛁ = tailwindcss logo ;) -->

<div class="⛁">
...

<!-- or maybe we could also show the number of classes as well? -->
<!-- [] = nice gray background -->

<div class="[⛁7]">
...
```

bradlc:
Hey @mtbaqer, great work on this, the little Tailwind logo is a nice touch! 👏 I think that HTML attribute folding is a really cool idea, and one that a lot of Tailwind users would appreciate for those long `class` attributes!

I think there's a couple of reasons that we wouldn't want to add this to Tailwind CSS IntelliSense though:

1. I still don't think it's possible to build an inline folding extension that we would feel comfortable publishing. In particular existing implementations don't work well when word wrap is enabled. Maybe one day it will be possible (maybe it is now? I would love to be proven wrong), or we'll get first-party support for inline code folding 🤞
2. Implementation issues aside, this doesn't feel like a feature that belongs in Tailwind CSS IntelliSense. Attribute folding (even `class` attribute folding specifically) is not a Tailwind-specific feature, so it feels weird to me to include it in a Tailwind-specific extension. I think a standalone extension makes more sense, like `vscode-inline-fold`.

For those reasons I'm going to close this, I hope that's understandable 🙏

PR #723, opened 2023-03-04
Fix InvalidTailwindDirective in notSemicolonLanguages with CRLF file endings

TheRealSyler:
Just adds `(\r)?` to the find tailwind directive regex so that it recognizes `\r\n` as the end of line, this was the error that i got:
```
'utilities
' is not a valid value. Did you mean 'utilities'?(invalidTailwindDirective)
```

### steps to reproduce 

1. go into any tailwind app
2. create `app.sass`
3. paste `@tailwind utilities`
4. change file endings to CRLF
5. now you should see the warning


thecrypticace:
Thanks! Made one tweak as `[^\r\n]` is what you want here. The regex `[^(\r)?\n]` would match any character that is not: `\r`, `\n`, `(`, `)`, or `?` when we just want to exclude either `\r` or `\n`.

PR #725, opened 2023-03-10
Support `InitializeParams.rootUri`

### LINKED ISSUE #724: Unable to initialise language server due to missing rootPath

v0.0.10 of the npm package (via `npm install -g tailwindcss-language-server`)

v3.2.7

npm

macOS

**Tailwind config**

```js
module.exports = {
  content: ['./index.html'],
  plugins: [],
}
```

**~~VS Code~~ Vim settings**

I'm using Vim 9.0.472 and the [vim-lsc](https://github.com/natebosch/vim-lsc) plugin.  My vimrc config is:

```vim
let g:lsc_server_commands = {'html': 'tailwindcss-language-server --stdio'}
```

**Describe your issue**

I have a single `index.html` file in my Tailwind project directory.  Tailwind's CLI works fine because it detects my classes and generates the stylesheet as expected.

I want to be able to (a) autocomplete Tailwind's classes and (b) hover over a class to see the actual CSS it corresponds to.

I assume the language server is the best way to achieve this?  Or is there a way to generate a stylesheet with every possible Tailwind CSS class, which I could then use a source for autocompletions?

Anyway, when I open my index.html in vim, I see the error message `No workspace folders found, not initializing.`

I believe this error comes from here: https://github.com/tailwindlabs/tailwindcss-intellisense/blob/3aafffefdab2deafdfe6c3b4acbbe9cd1ebf6d04/packages/tailwindcss-language-server/src/server.ts#L1610-L1613

Logging shows me that vim-lsc sends this to the language server:

```json
{"method":"initialize","jsonrpc":"2.0","id":1,"params":{"rootUri":"file:///path/to/my/project","capabilities":{"workspace":{"configuration":true,"applyEdit":true},"textDocument":{"completion":{"completionItem":{"snippetSupport":false,"tagSupport":{"valueSet":[1]},"deprecatedSupport":true}},"synchronization":{"willSaveWaitUntil":false,"willSave":false,"didSave":false},"codeAction":{"codeActionLiteralSupport":{"codeActionKind":{"valueSet":["quickfix","refactor","source"]}}},"signatureHelp":{"dynamicRegistration":false},"hover":{"contentFormat":["plaintext","markdown"]},"definition":{"dynamicRegistration":false}}},"clientInfo":{"name":"vim-lsc"},"processId":62896,"trace":"off"}}Content-Length: 52

{"method":"initialized","jsonrpc":"2.0","params":{}}Content-Length: 20281

{"method":"textDocument/didOpen","jsonrpc":"2.0","params":{"textDocument":{"uri":"file:///path/to/my/project/index.html","version":1,"languageId":"html","text":"..."}}}Content-Length: 44

{"method":"shutdown","jsonrpc":"2.0","id":2}Content-Length: 33

{"method":"exit","jsonrpc":"2.0"}
```

And gets these responses:

```json
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"No workspace folders found, not initializing."}}Content-Length: 276

{"jsonrpc":"2.0","id":1,"result":{"capabilities":{"textDocumentSync":1,"hoverProvider":true,"colorProvider":true,"codeActionProvider":true,"documentLinkProvider":{},"completionProvider":{"resolveProvider":true,"triggerCharacters":["\"","'","`"," ",".","(","[","!","/",":"]}}}}Content-Length: 38

{"jsonrpc":"2.0","id":2,"result":null}
```

Do you think the problem is with what Vim is sending to the language server?  Should it be including a `rootPath` parameter somewhere in the initial message?  (I tried including one inside the `params`, with the same value as the `rootUri` parameter, but then I got an `Unhandled exception: No such file or directory` error messsage.)

I'd be grateful for any suggestions – thanks!

bradlc:
Fixes #724

### DIFF
--- a/packages/tailwindcss-language-server/src/server.ts
+++ b/packages/tailwindcss-language-server/src/server.ts
@@ -1609,7 +1609,6 @@ class TW {
 
     let base: string
     if (this.initializeParams.rootUri) {
-      console.log('here!!')
       base = URI.parse(this.initializeParams.rootUri).fsPath
     } else if (this.initializeParams.rootPath) {
       base = normalizeFileNameToFsPath(this.initializeParams.rootPath)


airblade:
Thank you!

PR #735, opened 2023-03-15
Use @types/culori

remcohaszing:
This removes the need to define types here. Also this allows third parties to use this package without having to declare type definitions for culori.


bradlc:
Thank you!

PR #738, opened 2023-03-20
Support TS/ESM Tailwind config files

bradlc:
This PR adds support for TypeScript and ESM Tailwind config files when using a version of `tailwindcss` that supports these (currently `tailwindcss@insiders`, since https://github.com/tailwindlabs/tailwindcss/pull/10785)


polyzen:
@bradlc could you please cut a new language-server release, to include this and #739?

PR #747, opened 2023-03-27
Fix `theme` helper handling when specifying default value

### LINKED ISSUE #734: `invalidConfigPath` error when passing default value for `theme()` function

`v0.9.8`

`v3.2.7`

`npm`

`macOS`

**Tailwind config**

```js
module.exports = {
  content: ['./src/**/*.{js,jsx,ts,tsx}'],
  theme: {
    extend: {},
  },
  plugins: [],
}
```

**VS Code settings**

```json
{
  "tailwindCSS.emmetCompletions": true
}
```

**Reproduction URL**

https://github.com/javimsevilla/tailwindcss-intellisense-theme-issue

**Describe your issue**

I want to pass a default value for the TailwindCSS `theme()` function as a second argument, but tailwindcss-intellisense shows that this is an error, but it is not, the code works as expected.

image

It doesn't matter what I pass as arguments, it always shows the error if I pass a second argument.

image

As a workaround, I can disable the `"tailwindCSS.lint.invalidConfigPath"` property in VS Code settings:

```json
{
  "tailwindCSS.lint.invalidConfigPath": "ignore"
}
```
But, I think this is a bug because it's not the expected behavior.

bradlc:
Fixes #734

### DIFF
--- a/packages/tailwindcss-language-service/src/util/find.ts
+++ b/packages/tailwindcss-language-service/src/util/find.ts
@@ -348,6 +348,22 @@ export function findHelperFunctionsInDocument(
   )
 }
 
+function getFirstCommaIndex(str: string): number | null {
+  let quoteChar: string | undefined
+  for (let i = 0; i < str.length; i++) {
+    let char = str[i]
+    if (char === ',' && !quoteChar) {
+      return i
+    }
+    if (!quoteChar && (char === '"' || char === "'")) {
+      quoteChar = char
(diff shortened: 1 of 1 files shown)


simonhagger:
Please see this note on 734:
https://github.com/tailwindlabs/tailwindcss-intellisense/issues/734#issuecomment-2394992343
This fix does not stop interference with third party implementations - just for reference.

PR #779, opened 2023-05-09
Improved performance

MaximilianLloyd:
Improved performance by avoid resizing matches array. This seems to provide a quite healthy performance improvement across the board. 

I did some performance tracking with `node:perf_hooks`
image

The first block is without my modifications, just tracking the performance. The second is with my changes. I tested it by modifying the className with exactly the same values. With my changes it seems that around 100-150~ms are sliced off on many calls.

### DIFF
--- a/packages/tailwindcss-language-service/src/util/find.ts
+++ b/packages/tailwindcss-language-service/src/util/find.ts
@@ -14,12 +14,18 @@ import { getTextWithoutComments } from './doc'
 import { isSemicolonlessCssLanguage } from './languages'
 
 export function findAll(re: RegExp, str: string): RegExpMatchArray[] {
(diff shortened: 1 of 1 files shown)


richiedevs:
Awesome work man!

bradlc:
Hey @MaximilianLloyd, thanks for the PR. On this occasion I'm going to leave this as it is for now, because I'm not seeing any performance improvement on my end and I don't want to make changes like this if there isn't a significant benefit. I hope that's understandable!

As a side note, I believe the numbers you shared are in milliseconds, not seconds, so the improvement you are seeing is 0.1ms, not 100ms.

MaximilianLloyd:
> 

My bad, sorry for the inconvenience 😊

PR #781, opened 2023-05-12
Migrate from deprecated text document

MaximilianLloyd:
Did some spring cleaning and replaced deprecated instances of TextDocument in the codebase. Verified that the LSP builds and works as normal.


bradlc:
Thank you!

Issue #783, opened 2023-05-17
Extension Not Starting


albertreed:
**What version of Tailwind CSS IntelliSense are you using?**

0.9.11

**What version of Tailwind CSS are you using?**

3.3.2

**What package manager are you using?**

yarn

**What operating system are you using?**

Mac

**Tailwind config**

```js
/** @type {import('tailwindcss').Config} */
export default {
  content: ["./index.html", "./src/**/*.{js,ts,jsx,tsx}"],
  theme: {
    extend: {},
  },
  plugins: [],
  // I have no idea why this changed everything to be like it was with tachyons...but here we are.
  corePlugins: ["lineHeight"],
};

```

**VS Code settings**

Workspace
```json
{
  "folders": [
    {
      "path": "."
    }
  ],
  "settings": {
    "editor.formatOnSave": true,
    "eslint.workingDirectories": ["./client", "./server"],
    "jest.pathToJest": "${workspaceFolder}/server/node_modules/.bin/jest",
    "jest.pathToConfig": "${workspaceFolder}/server/jest.config.js",
    "jest.rootPath": "${workspaceFolder}/server/",
    "files.exclude": {
      "${workspaceFolder}/shared": true,
      "client/src/index.css": true,
      "moderation-dashboard/**": true,
      "node_modules/**": true,
      "steam-sdk/**": true
    },
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "search.exclude": {
      "**/__generated__": true,
      "**/*.lock": true,
      "**/dist": true,
      "**/generatedTypes": true
    },
    "terminal.integrated.tabs.enabled": false,
    "tailwindCSS.includeLanguages": {
      "plaintext": "javascript"
    },
    "tailwindCSS.codeActions": true,
    "tailwindCSS.hovers": true
  },
  "extensions": {
    "recommendations": [
      "dbaeumer.vscode-eslint",
      "esbenp.prettier-vscode",
      "redhat.vscode-yaml",
      "forbeslindesay.vscode-sql-template-literal"
    ]
  }
}

```

User
```json
{
  "editor.minimap.enabled": false,
  "html.autoClosingTags": false,
  "typescript.autoClosingTags": false,
  "diffEditor.ignoreTrimWhitespace": true,
  "editor.autoClosingQuotes": "never",
  "editor.autoClosingBrackets": "never",
  "jestrunner.jestCommand": "docker-compose run server test",
  "editor.formatOnSave": true,
[...]

PR #787, opened 2023-05-25
Fix initialisation when using `tailwindcss@^0`

bradlc:
Reference: https://github.com/tailwindlabs/tailwindcss-intellisense/issues/783#issuecomment-1562535023


Mr-KayJayDee:
Do i need to wait for an extension update or can i use this now ? It does not seem fixed on my side and i don't know how to package extensiosn from code.

PR #851, opened 2023-08-31
Add "Sort Selection" command

bradlc:
This PR adds a new "Sort Selection" command to the extension. When activated this command will sort the selection(s) in the [same way](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted) that [`prettier-plugin-tailwindcss`](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) sorts lists of CSS classes.

This useful when you want to sort a list of classes that `prettier-plugin-tailwindcss` does not handle, e.g.

```js
const myVar = 'these classes here'
```

https://github.com/tailwindlabs/tailwindcss-intellisense/assets/2615508/2b141fac-b5ed-4537-b8af-28a946548f43

The command is only available when:

- The editor has one or more active text selections
- There is an active Tailwind project for the current active file
- The Tailwind version for the active Tailwind project is at least `3.0.0`
- The active file is not excluded (via `tailwindCSS.files.exclude`)


Benbinbin:
really need "sort classes name" as the built-in feature the extension, maybe take a further step to make this feature as "sort on save" will be more convenient

PR #867, opened 2023-10-14
feat: add glimmer langs

IgnaceMaes:
Adds support for `.gjs` and `.gts` file formats. Glimmer `<template>` tag components are the new format which power Ember.js. References:

- RFC: https://rfcs.emberjs.com/id/0779-first-class-component-templates
- Component implementation: https://github.com/ember-template-imports/ember-template-imports

Local test:

image


thecrypticace:
Thanks — tested this and it works as expected! Appreciate it.

This should be in the pre-release extension soon-ish (~10 minutes I think?)

PR #881, opened 2023-11-22
Simplify hovers

maddymeows:
Enables hovers to replace the class name with a simple `&` to better help understand what the selector is doing.

Sometimes it is necessary to write code that gets as complex as `[&[aria-selected=true]:has(+_:not([aria-selected=true]))::before]:border-purple-500`. I often reach to hover over the class to make sure it is doing what I intended. Unfortunately the current implementation's hover puts out this code:

```css
.\[\&\[aria-selected\=true\]\:has\(\+_\:not\(\[aria-selected\=true\]\)\)\:\:before\]\:border-purple-500[aria-selected=true]:has(+ :not([aria-selected=true]))::before {
  --tw-border-opacity: 1;
  border-color: rgb(168 85 247 / var(--tw-border-opacity));
}
```

I would not blame you if you had no idea what this selector was doing, and the noise gets amplified by the word wrapping that VS code does. It takes quite a lot of time to precisely figure out which part of this selector is relevant. With my changes the hover would instead show the following:

```css
&[aria-selected=true]:has(+ :not([aria-selected=true]))::before {
  --tw-border-opacity: 1;
  border-color: rgb(168 85 247 / var(--tw-border-opacity));
}
```

I've opted to make this default behaviour as I believe this is the better experience to provide by default.

### DIFF
--- a/packages/tailwindcss-language-server/src/server.ts
+++ b/packages/tailwindcss-language-server/src/server.ts
@@ -251,7 +251,7 @@ async function getConfiguration(uri?: string) {
         classAttributes: ['class', 'className', 'ngClass'],
         codeActions: true,
         hovers: true,
-        simplifyHovers: false,
+        simplifyHovers: true,
         suggestions: true,
         validate: true,
         colorDecorators: true,
--- a/packages/tailwindcss-language-server/src/server.ts
+++ b/packages/tailwindcss-language-server/src/server.ts
@@ -251,6 +251,7 @@ async function getConfiguration(uri?: string) {
         classAttributes: ['class', 'className', 'ngClass'],
         codeActions: true,
         hovers: true,
+        simplifyHovers: false,
         suggestions: true,
         validate: true,
         colorDecorators: true,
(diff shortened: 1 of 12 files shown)


thecrypticace:
I really like the idea in a general sense but I think we want to approach this in a different way. In v4 we use native CSS nesting syntax which, in intellisense, helps with understanding what each variant applied to a class is doing.

For example, instead of using one long arbitrary variant you can write the class you have like so in v4:
```
aria-selected:[&:has(+_:not([aria-selected=true]))]:before:border-purple-500
```

And when hovering over it you'll get this:
Screenshot 2024-06-04 at 11 36 26

And each variant is separated here so it's easier to see what is happening. Given that, I think if we were to do anything here it would be to add a feature that tries to convert the flattened CSS to nested CSS. And we won't need to do this in v4 because it'll just work that way by default.

I'm gonna close this but I'll make a note internally to investigate what it would take to "unflatten" CSS for earlier Tailwind versions and see if we want to offer an option for that if it's ends up being workable.

PR #890, opened 2023-12-20
Support Astro's `class:list` attribute by default

yuheiy:
In Astro, `class:list={...}` takes an array of class values.

```astro
<!-- This -->
<span class:list={[ 'hello goodbye', { world: true }, [ 'friend' ] ]} />
<!-- Becomes -->
<span class="hello goodbye world friend"></span>
```

https://docs.astro.build/en/reference/directives-reference/#classlist

Including `class:list` in the default `tailwindCSS.classAttributes` should make it more useful!


thecrypticace:
Thanks!

PR #930, opened 2024-03-21
Fix hovers and conflicts in Vue `<style lang="sass">` blocks

### LINKED ISSUE #895: CSS conflicts in `@apply` don't handle `.sass` syntax

v1.85.1

v0.10.4

v3.4.0

npm

Windows

**Tailwind config**

```js
export default {
  content: [
    "./index.html",
    "./src/**/*.{vue,js,ts,jsx,tsx}",
  ],
  theme: {
    extend: {},
  },
  plugins: [],
}
```

**VS Code settings**

```json
// Paste your VS Code settings in JSON format here
```

**Reproduction URL**

[cssConflictBug](https://github.com/FragsterAt/cssConflictBug)

**Describe your issue**

When using sass in vue sfc components with `@apply` there is false positive diagostic cssConflict fires in some cases:
изображение

изображение

it can be not only comment, but any string with `}` bracket
bug in https://github.com/FragsterAt/cssConflictBug/blob/main/src/App.vue file

thecrypticace:
We weren't parsing these blocks as Sass (or any semicolon-less language) and we were instead assuming CSS syntax. Because of this we wouldn't pick up conflicts when we should and sometimes would pick up conflicts when we shouldn't.

This PR fixes this by handling language boundaries on hover and threading language information from `<style lang="…">` blocks.

Fixes #895


RobinMalfait:
Do you think it's a lot of work to setup CI for tests? I can't figure out if these tests are expected to fail or not:

image

RobinMalfait:
Ah figured it out, I had to manually go into each `tests/fixtures/v{1,4}` and run `npm install`. I think this is something we should improve in a future PR to make sure that all those sub folders are part of the workspace if possible.

packages/tailwindcss-language-service/src/util/find.ts
```diff
@@ -97,9 +97,10 @@ export function findClassListsInCssRange(
   state: State,
   doc: TextDocument,
   range?: Range,
+  lang?: string,
 ): DocumentClassList[] {
   const text = getTextWithoutComments(doc, 'css', range)
-  let regex = isSemicolonlessCssLanguage(doc.languageId, state.editor?.userLanguages)
+  let regex = isSemicolonlessCssLanguage(lang ?? doc.languageId, state.editor?.userLanguages)
```
RobinMalfait, line 103: Just a question, in what scenario would `lang` and `doc.languageId` be different?

is that when you have `<style lang="sass">` in an html or vue file where `lang` will be `sass` and `doc.languageId` would be `html`?

RobinMalfait (approved):

thecrypticace:
> Ah figured it out, I had to manually go into each `tests/fixtures/v{1,4}` and run `npm install`. I think this is something we should improve in a future PR to make sure that all those sub folders are part of the workspace if possible.

We can't make them a part of the normal project workspace because of dependency hoisting, unfortunately. If there's a way to disable hoisting but still use workspaces I'd use them for all of these.

packages/tailwindcss-language-service/src/util/find.ts
thecrypticace: vue file with a <style> block or a <script> block. Also HTML file with a <script> block. `doc` always refers to the whole file/document rather than just a portion of the file.

thecrypticace:
I need to fix the test:prepare script though, good call. I'll do that in a separate PR.

packages/tailwindcss-language-service/src/util/find.ts
RobinMalfait: Gotcha yep that's what I meant and thought. Thanks for verifying!

PR #945, opened 2024-04-03
Fix errors thrown by detecting content files with oxide (#942)

hampustagerud:
I had the same problem as described in #942, which seem to have been closed preemptively 🙂 By adding oxide as a devDependency and then checking the import with a type guard we get an error that explains the issue:

Screenshot 2024-04-03 at 21 43 10

So by ignoring the globs when creating the DocumentSelectors, `normalizePath` no longer throws and the server starts up properly in a fresh v4 setup 👍

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -725,6 +725,187 @@
         "tailwindcss": ">=2.0.0 || >=3.0.0 || >=3.0.0-alpha.1"
       }
     },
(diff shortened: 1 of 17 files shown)


thecrypticace:
Ah right the types changed here at some point during development. I'll get this fixed up and merged in today.

thecrypticace:
Thanks!

PR #947, opened 2024-04-05
feat: add VS Code configuration defaults

eltigerchino:
The [recommended VS code settings](https://github.com/tailwindlabs/tailwindcss-intellisense#recommended-vs-code-settings) can be provided by default when the extension is installed using the [configuration defaults](https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults) contribution point in the extension `package.json`. This would mean less setup for users who install the extension and benefits people who may have missed reading the README first (sadly I was one of those people).

### DIFF
--- a/packages/vscode-tailwindcss/package.json
+++ b/packages/vscode-tailwindcss/package.json
@@ -323,6 +323,14 @@
           "markdownDescription": "Enable the Node.js inspector agent for the language server and listen on the specified port."
         }
       }
+    },
+    "configurationDefaults": {
+      "files.associations": {
+        "*.css": "tailwindcss"
+      },
+      "editor.quickSuggestions": {
+        "strings": "on"
+      }
     }
   },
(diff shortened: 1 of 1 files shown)


thecrypticace:
Hey! So, I've been thinking about this one for a while and I was worried that adding these defaults might cause a bit of surprise for people who have the extension installed *but* are not using Tailwind CSS in some of their projects. As a potential solution to this, I implemented an "auto-switch" functionality for CSS files in detected Tailwind CSS projects which automatically changed the language mode of CSS files over to "Tailwind CSS". Turns out though this was still quite unexpected for some users even when only operating within detected Tailwind CSS projects.

Given that I'm going to close this as I think it'll introduce the same problem but on a potentially larger scale. I'd prefer that the user's who need to make these changes to the config do so themselves. Thanks anyway for the suggestion! 🙏

PR #957, opened 2024-04-23
Use one server to handle all folders in a workspace

### LINKED ISSUE #921: Intellisense does not activate when the directory is included in a workspace but not the first directory.
**Describe your issue**

When using this file path for my VS workspace, I run into an issue where the extension does not recognize the config file.

	"folders": [
		{
			"path": "src"
		},
		{
			"path": "../../../Components"
		},
		{
			"path": "."
		}
		
When trying to use . at the top path it works however my live server plugin then breaks, I've been fiddling with both extensions to no avail. Is there a better way to setup my directory so they both work or is there any setting options to make the extension to check under all the files?

thecrypticace:
Previously, if your workspace had multiple folders, we would boot a separate server per workspace folder. This was because our server didn't support workspace folders. However, it now does which means we can make use of this feature.

This change should help reduce CPU and memory usage in any workspace making use of multiple folders. This will also solve an issue where Intellisense would not load when a workspace that is open is a subdirectory of another workspace, the subdirectory folder was listed first, and that subdirectory didn't contain a tailwind config or tailwind-related CSS file.

Fixes #921

### DIFF
--- a/packages/vscode-tailwindcss/src/extension.ts
+++ b/packages/vscode-tailwindcss/src/extension.ts
@@ -17,7 +17,11 @@ import {
   Range,
   RelativePattern,
 } from 'vscode'
-import type { DocumentFilter, LanguageClientOptions, ServerOptions } from 'vscode-languageclient/node'
+import type {
+  DocumentFilter,
+  LanguageClientOptions,
+  ServerOptions,
+} from 'vscode-languageclient/node'
 import {
   LanguageClient,
   TransportKind,
@@ -295,7 +299,6 @@ export async function activate(context: ExtensionContext) {
       let needsReboot = folders.some((folder) => {
         return (
           event.affectsConfiguration('tailwindCSS.experimental.configFile', folder) ||
-
           // TODO: Only reboot if the MAPPING changed instead of just the languages
           // e.g. "plaintext" already exists but you change it from "html" to "css"
           // TODO: This should not cause a reboot of the server but should instead
@@ -383,8 +386,11 @@ export async function activate(context: ExtensionContext) {
       module = context.asAbsolutePath(prod)
     } catch (_) {}
 
-    let workspaceFile = Workspace.workspaceFile?.scheme === 'file' ? Workspace.workspaceFile : undefined
-    let inspectPort = Workspace.getConfiguration('tailwindCSS', workspaceFile).get<number | null>('inspectPort') ?? null
+    let workspaceFile =
+      Workspace.workspaceFile?.scheme === 'file' ? Workspace.workspaceFile : undefined
+    let inspectPort =
+      Workspace.getConfiguration('tailwindCSS', workspaceFile).get<number | null>('inspectPort') ??
+      null
 
     let serverOptions: ServerOptions = {
       run: {
@@ -419,11 +425,7 @@ export async function activate(context: ExtensionContext) {
           let selections = editor.selections
           let edits = result.additionalTextEdits || []
 
-          if (
-            selections.length <= 1 ||
-            edits.length === 0 ||
-            result['data'] !== 'variant'
-          ) {
+          if (selections.length <= 1 || edits.length === 0 || result['data'] !== 'variant') {
             return result
           }
 
@@ -491,9 +493,7 @@ export async function activate(context: ExtensionContext) {
         workspace: {
           configuration: (params) => {
             return params.items.map(({ section, scopeUri }) => {
-              let scope: ConfigurationScope | null = scopeUri
-                ? Uri.parse(scopeUri)
-                : null
+              let scope: ConfigurationScope | null = scopeUri ? Uri.parse(scopeUri) : null
 
               let settings = Workspace.getConfiguration(section, scope)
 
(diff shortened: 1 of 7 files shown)


thecrypticace:
@alexander-doroshko I'm landing a change here that will have the server ask for "user languages" from the config rather than from initialization options. We have an existing setting for this that we use populate the initialization options in our extension: `tailwindCSS.includedLanguages`.

Would this cause any problems with jet brains stuff? Right now it'll still fall back to initialization options so it _shouldn't_ break if none are defined in the config but it's likely I'll remove that in the near future.

alexander-doroshko:
> jet brains

JetBrains :)

> ask for "user languages" from the config

Do you mean the server will ask
```json
  "method": "workspace/configuration",
  "params": {
    "items": [
      {
        "section": "tailwindCSS"
      }
    ]
  }
```
and the IDE is expected to answer with
```json
  "result": [
    {
      "includeLanguages": {
        "ftl": "html",
        "jinja": "html",
        "jinja2": "html",
        "smarty": "html",
        "tmpl": "gohtml",
        "cshtml": "html",
        "vbhtml": "html",
        "razor": "html"
      },
      "files": {
        "exclude": [
          "**/.git/**",
          "**/node_modules/**",
          "**/.hg/**",
          "**/.svn/**"
        ]
      },
      "emmetCompletions": false,
      "classAttributes": [
        "class",
        "className",
        "ngClass"
      ],
      "colorDecorators": false,
      "showPixelEquivalents": true,
      "rootFontSize": 16,
      "hovers": true,
      "suggestions": true,
      "codeActions": true,
      "validate": true,
      "lint": {
        "invalidScreen": "error",
        "invalidVariant": "error",
        "invalidTailwindDirective": "error",
        "invalidApply": "error",
        "invalidConfigPath": "error",
        "cssConflict": "warning",
        "recommendedVariantOrder": "warning"
      },
      "experimental": {
        "classRegex": []
      }
    }
  ]
```
?

alexander-doroshko:
Cool, thanks, this is what our Tailwind CSS plugin already does.

Anyway, we'll check everything once again when we are updating the bundled Tailwind CSS LSP server. Please ping me somehow when you publish a new version to https://www.npmjs.com/package/@tailwindcss/language-server. For example, by email `alexander.doroshko at jetbrains.com`.

thecrypticace:
Will do!

packages/vscode-tailwindcss/src/extension.ts
```diff
@@ -297,147 +290,129 @@ export async function activate(context: ExtensionContext) {
...
-        }
+      if (!needsReboot) {
+        return
+      }
 
-        if (reboot && client) {
-          toReboot.add(folder)
-        }
-      })
+      // IDEA: We might be able to service requests during a reboot by booting the
```
RobinMalfait, line 311: Is this todo/idea still relevant, or do we want to tackle this in a separate PR?

packages/vscode-tailwindcss/src/extension.ts
```diff
@@ -684,24 +623,20 @@ export async function activate(context: ExtensionContext) {
...
-      for (let folder of event.removed) {
-        let client = clients.get(folder.uri.toString())
-        if (client) {
-          searchedFolders.delete(folder.uri.toString())
-          clients.delete(folder.uri.toString())
-          client.stop()
-        }
-      }
+    Workspace.onDidChangeWorkspaceFolders(async () => {
+      if (Workspace.workspaceFolders?.length ?? 0 > 0) return
```
RobinMalfait, line 640: I think you want `()` here because of operator precedence, it just happens to work because numbers are truthy _and_ we don't have negative lengths _and_ `0` is falsey.

```suggestion
      if ((Workspace.workspaceFolders?.length ?? 0) > 0) return
```

Because right now it's doing `expression ?? (0 > 0)`:
```
> 20 ?? 0 > 0
20
> 20 ?? (0 > 0)
20
> (20 ?? 0) > 0
true
```

packages/vscode-tailwindcss/src/extension.ts
```diff
@@ -301,363 +289,216 @@ export async function activate(context: ExtensionContext) {
...
+          let selections = editor.selections
+          let edits = result.additionalTextEdits || []
+
           if (
-            result['data'] === 'variant' &&
-            selections.length > 1 &&
-            result.additionalTextEdits?.length > 0
+            selections.length <= 1 ||
+            edits.length === 0 ||
+            // @ts-expect-error: use of private API / implementation details of our language server
```
RobinMalfait, line 428: This gives me an error in my editor and while creating a build that it is unused. I guess this is not an actual error anymore 🤔

RobinMalfait: (Deleting this, will result in Prettier changes)

RobinMalfait: Oh actually, if you change `result['data']` to `result.data` it _does_ show up as an error and the `// @ts-expect-error` is required.

RobinMalfait (commented):
- When running `prettier --write .` I do get some changes in files

Built a local vsix file and everything still seems to work!

packages/vscode-tailwindcss/src/extension.ts
thecrypticace: VScode shows this as an error but for some reason Zed does not 🤷‍♂️

packages/vscode-tailwindcss/src/extension.ts
thecrypticace: It is but it's for something to be done later. I'll drop this and add a card for it.

packages/vscode-tailwindcss/src/extension.ts
thecrypticace: Fixed. I pulled it out to a separate variable. Feels cleaner.

RobinMalfait (approved):

PR #970, opened 2024-05-21
feat: remove is-builtin-module dependency

43081j:
Node provides a list of built ins via `builtinModules`. We can test against this rather than including an extra (redundant) dependency.

In future, if we introduce a version constraint of node `>=16`, we can use `isBuiltin` which node provides:

```ts
import {isBuiltin} from 'node:module';

isBuiltin('fs'); // true
```

### DIFF
--- a/packages/tailwindcss-language-server/src/lib/env.ts
+++ b/packages/tailwindcss-language-server/src/lib/env.ts
@@ -1,15 +1,16 @@
 import Module from 'module'
 import * as path from 'path'
 import resolveFrom from '../util/resolveFrom'
-import isBuiltinModule from 'is-builtin-module'
+import {builtinModules} from 'module'
 
 process.env.TAILWIND_MODE = 'build'
 process.env.TAILWIND_DISABLE_TOUCH = 'true'
 
 let oldResolveFilename = (Module as any)._resolveFilename
+const nodeModulePrefix = /^node:/i;
 
 ;(Module as any)._resolveFilename = (id: any, parent: any) => {
-  if (typeof id === 'string' && isBuiltinModule(id)) {
+  if (typeof id === 'string' && builtinModules.includes(id.replace(nodeModulePrefix, ''))) {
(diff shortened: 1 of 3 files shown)


packages/tailwindcss-language-server/src/lib/env.ts
```diff
@@ -1,15 +1,16 @@
...
 
 process.env.TAILWIND_MODE = 'build'
 process.env.TAILWIND_DISABLE_TOUCH = 'true'
 
 let oldResolveFilename = (Module as any)._resolveFilename
+const nodeModulePrefix = /^node:/i;
 
 ;(Module as any)._resolveFilename = (id: any, parent: any) => {
-  if (typeof id === 'string' && isBuiltinModule(id)) {
+  if (typeof id === 'string' && builtinModules.includes(id.replace(nodeModulePrefix, ''))) {
```
thecrypticace, line 13: Is there a benefit to using `builtinModules.includes` rather than `isBuiltinModule`?

thecrypticace: `builtinModules` seems to be a list of modules that can be required with or without the node prefix and `node:*` bypasses the require cache so it's probably the same result either way.

43081j: the benefit is that we're dropping a dependency here and using the built in set from node itself

if you bump your `engines` constraint to `>=16`, we could just use `isBuiltin` from node instead of an `includes`

thecrypticace: Oh, I see. `builtinModules` is an older API than `isBuiltinModule` — missed that in the Node docs. 👍

I kinda doubt the language server will run on Node < 16 anyway but it should probably be bumped to at least Node 18 given that 16 is well past its EOL and 18 is only in maintenance mode at this point anyway. Will do that soon.

thecrypticace:
Thanks!

PR #907, opened 2024-02-14
fix: bump `enhanced-resolve` for upstream fix

llllvvuu:
EDIT: Sorry, I just noticed https://github.com/webpack/enhanced-resolve/pull/301 still hasn't been merged. In that case there would be a need to rebase enhanced-resolve-301 instead of this.

This pulls in https://github.com/webpack/enhanced-resolve/pull/353.

Otherwise, the following is broken (https://github.com/webpack/enhanced-resolve/issues/352):
```javascript
module.exports = require('@acme/ui/tailwind.config.cjs')
```

where `@acme/ui` has:
```json
"exports": {
  "./*.cjs": "./*.cjs",
},
```

and requires the following workaround
```json
"exports": {
  "./tailwind.config.cjs": "./tailwind.config.cjs",
  "./*.cjs": "./*.cjs",
},
```

Bumping the package fixes this. The `tailwindcss-language-server` code has a comment referencing https://github.com/webpack/enhanced-resolve/issues/282, but that's been fixed upstream so I'm not sure if it's still a blocker for this.

PR #971, opened 2024-05-21
chore: bump enhanced-resolve version

43081j:
Builds on #907 but bumps the version of enhanced-resolve to latest.

Also drops a `replace` which should now be solved inside enhanced-resolve itself.

tests seem to pass but i'm not aware of a specific one to test what the `replace` was fixing/working around. if someone can point me in the right direction, im happy to add one


thecrypticace:
I tried to create a project similar to what was being fixed originally but it doesn't work in the existing extension 🙃 

This doesn't seem to break other things though so I'm going to roll with it for now and will keep an eye out for PnP related issues. Thanks!

43081j:
thanks for taking a look 🙏 

if any issues do come up, give me a shout and ill be happy to help

PR #981, opened 2024-06-18
Fallback to global settings when requests for folder settings fail

thecrypticace:
Zed seems to return an error from their LSP: `invalid value: string “…”, expected relative URL without a base`. That seems like a bug in Zed (will file an issue today probably) but it seems possible that other LSP implementations would have the same problem.

### DIFF
--- a/packages/tailwindcss-language-server/src/tw.ts
+++ b/packages/tailwindcss-language-server/src/tw.ts
@@ -177,7 +177,7 @@ export class TW {
       userLanguages = folderSettings.tailwindCSS.includeLanguages
     } catch (error) {
       console.error(
-        'Unable to get get settings for workspace folder. Using global settings instead.',
+        'Unable to get the settings for workspace folder. Using global settings instead.',
         error,
       )
     }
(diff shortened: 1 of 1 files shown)


packages/tailwindcss-language-server/src/tw.ts
```diff
@@ -164,8 +170,18 @@ export class TW {
...
-    let folderSettings = await this.settingsCache.get(base)
-    let userLanguages = folderSettings.tailwindCSS.includeLanguages
+    let userLanguages = globalSettings.tailwindCSS.includeLanguages
+
+    try {
+      let folderSettings = await this.settingsCache.get(base)
+      userLanguages = folderSettings.tailwindCSS.includeLanguages
+    } catch (error) {
+      console.error(
+        'Unable to get get settings for workspace folder. Using global settings instead.',
```
RobinMalfait, line 180:
```suggestion
        'Unable to get the settings for workspace folder. Using global settings instead.',
```

RobinMalfait (approved):

PR #982, opened 2024-06-18
Pass URI to configuration call not a file path

thecrypticace:
Not a Zed bug — finally got the init messages to show in Zed. We're passing a path and not a `file://` URI which is a bug on our side. Also fixed another issue while I was in there related to settings updates and color decorators.


RobinMalfait (approved):
Works in Zed & Neovim 💪

PR #1011, opened 2024-07-03
feat(util/languages): add built-in support for `.svx` files

fzn0x:
`.svx` is an extension used by [MDsveX](https://github.com/pngwn/MDsveX).

Current workaround is to add `.svx` in `settings.json` to support tailwindcss intellisense:

```json
{
    "svx": "html"
  }
}
```

Screenshot 2024-07-04 001843

### DIFF
--- a/packages/tailwindcss-language-service/src/util/languages.ts
+++ b/packages/tailwindcss-language-service/src/util/languages.ts
@@ -33,6 +33,7 @@ export const htmlLanguages = [
   'razor',
   'slim',
   'surface',
+  'svx',
   'twig',
 ]
 


thecrypticace:
What VSCode extension are you using that adds the `svx` language? `svx` in that list and in `includeLanguages` is not a file extension but a language ID.

fzn0x:
> What VSCode extension are you using that adds the `svx` language? `svx` in that list and in `includeLanguages` is not a file extension but a language ID.

It is extension supported by [https://mdsvex.pngwn.io/docs](https://mdsvex.pngwn.io/docs), I'm not using any `.svx` VSCode extension. You need to configure MDsveX in your svelte kit to run `.svx`

fzn0x:
I'm just thinking the `htmlLanguages` is the right place to tell tailwindcss-intelisense it is the same as `.html`

thecrypticace:
My point was more that adding `svx` to `htmlLanguages` is not sufficient because that is _not_ a list of file extensions. `svx` would be the identifier associated with a language _added by another extension_.

Here's VSCode with this PR — notice how a `.svx` file still says "Plain Text" — meaning that you will not get completions:

Screenshot 2024-07-12 at 13 51 14

However, if you tell VSCode that `svx` is actually a markdown file you'll get completions:
Screenshot 2024-07-12 at 13 54 49

So I'm gonna close this PR.

PR #1012, opened 2024-07-04
fix: enable $ in the `classAttributes` setting

colelawrence:
We maintain backward compatibility by checking if the existing classAttributes setting is already escaping the dollar sign.

The current workaround for this is to simply escape the `$` manually, which is what I can do in the meantime.

2024-07-04 phosphor code-workspace — phosphor (Workspace) at 12 32PM@2x

### DIFF
--- a/packages/tailwindcss-language-service/src/util/find.ts
+++ b/packages/tailwindcss-language-service/src/util/find.ts
@@ -158,7 +158,13 @@ async function findCustomClassLists(
 }
 
 export function matchClassAttributes(text: string, attributes: string[]): RegExpMatchArray[] {
-  const attrs = attributes.filter((x) => typeof x === 'string').flatMap((a) => [a, `\\[${a}\\]`])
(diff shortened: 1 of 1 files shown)


thecrypticace:
Because this is fed directly to a regex I'm inclined to leave this as is. I think it's likely people have learned to use / are using regex syntax _on purpose_ here and changing this could be a breaking change for them. Especially with no way to "unescape" the `$`.

Escaping `$` would likely be less of an issue then say `?` from a backwards compatibility point of view. No matter what I think we'd want to be consistent about either escaping either all or no regex special characters and escaping all of them here would _definitely_ be a breaking change.

Given that I'm gonna close this but may think about a similar-ish change for a future version of Intellisense.

colelawrence:
@thecrypticace - if any unescaped $ is used with the current classAttribute, then that simply isn't possibly working for people.

What do you think about updating the documentation to explicitly state that these `classAttributes` accept RegExp strings? I believe that would help anyone in the future (like myself) understand how to fix this issue.

PR #1029, opened 2024-08-06
Migrate to `pnpm`

philipp-spiess:
To be consistent with the defaults of the Tailwind V4 repo, I figured it might make sense to also migrate the intellisense extension to use `pnpm` as the default version manager. 

This was mostly motivated by me doing a clean `npm install` in the workspace root but that causes a lot of packages to be missing within `packages/vscode-tailwindcss` (e.g. the `concurrently` module) so builds wouldn't _just work_. 😬


pnpm-lock.yaml
philipp-spiess, line 1: created via `pnpm import`

.vscode/tasks.json
```diff
@@ -2,7 +2,7 @@
   "version": "2.0.0",
   "tasks": [
     {
-      "type": "npm",
+      "type": "pnpm",
```
philipp-spiess, line 5: need to revert

thecrypticace:
From what I can tell in my testing the `workspace:*` deps in dev dependencies don't break the built VSIX file. So I think the best option here is to merge and see if the marketplace extension validation complains for insiders builds.

thecrypticace (approved):

package.json
```diff
@@ -1,17 +1,19 @@
...
   "private": true,
-  "workspaces": [
-    "packages/*"
-  ],
   "devDependencies": {
     "@npmcli/package-json": "^5.0.0",
+    "@types/culori": "^2.1.0",
+    "culori": "^4.0.1",
+    "esbuild": "^0.20.2",
+    "minimist": "^1.2.8",
```
thecrypticace, line 9: I don't really like that these deps have to be at the root package level but it's happening because the shared esbuild.mjs file is at the root level rather than having an exact copy inside each package.

philipp-spiess:
@thecrypticace thanks for fixing this up :)

PR #1030, opened 2024-08-07
Add support for `@source`

philipp-spiess:
This PR does multiple things:

1. It adds support for the `@plugin` and `@source` directive in the CSS language.
2. We will now scan V4 config files for eventually defined `@source` rules and respect them similarly how we handled custom content paths in V3.
3. Add support for the the Oxide API coming in Alpha 20 (https://github.com/tailwindlabs/tailwindcss/pull/14187)

For detecting the right content, we load the Oxide API installed in the user's Tailwind project. To do this in a backward compatible way, we now also load the `package.json` file of the installed Oxide version and support previous alpha releases for a limited amount of time.


packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -341,6 +342,9 @@ export class ProjectLocator {
     // Resolve real paths for all the files in the CSS import graph
     await Promise.all(imports.map((file) => file.resolveRealpaths()))
 
+    // Resolve all @source directives
+    await Promise.all(imports.map((file) => file.resolveSourceDirectives()))
```
philipp-spiess, line 346: @thecrypticace I moved this to the "resolve" step now like we discussed in our call last week.

packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -510,25 +519,37 @@ async function* contentSelectorsFromCssConfig(entry: ConfigEntry): AsyncIterable
...
 
     for (let file of files) {
       yield normalizePath(file)
     }
 
-    for (let { base, glob } of globs) {
+    for (let { base, pattern } of globs) {
       // Do not normalize the glob itself as it may contain escape sequences
-      yield normalizePath(base) + '/' + glob
+      yield normalizePath(base) + '/' + pattern
```
philipp-spiess, line 552: This is going to break older versions of oxide with the extension. Should we patch over it until we release a more stable version?

packages/tailwindcss-language-server/src/resolve-css-imports.ts
```diff
@@ -15,8 +18,129 @@ const resolveImports = postcss([
...
 ])
 
 export function resolveCssImports() {
   return resolveImports
 }
+
+const SINGLE_QUOTE = "'"
+const DOUBLE_QUOTE = '"'
+
+export default function fixRelativePathsPlugin(): Plugin {
```
thecrypticace, line 31: This should live in a separate file. Probably make a `src/css` folder or something and put resolve-css-imports.ts in there as well

packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -589,6 +611,29 @@ class FileEntry {
...
+        AtRule: {
+          source: ({ params }: AtRule) => {
+            if (params[0] !== '"' && params[0] !== "'") {
+              return
+            }
+            this.sources.push(params.slice(1, -1))
+          },
+        },
+      },
+    ]).process(this.content, { from: this.realpath })
```
thecrypticace, line 634: This makes me realize we're going to parse stuff with PostCSS probably a lot more than is necessary. There's no reason to change it right now but it's definitely something we should address in the future.

packages/tailwindcss-language-server/src/project-locator.ts
thecrypticace: Yes, we should. The first version should basically call the "modern" scanDir and then if it throws call `oxide.scanDir({ base, globs: true })` and map the globs so it's compatible.

packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -510,25 +519,37 @@ async function* contentSelectorsFromCssConfig(entry: ConfigEntry): AsyncIterable
...
+  sources: string[],
+): AsyncIterable<string> {
   try {
     let oxidePath = resolveFrom(path.dirname(base), '@tailwindcss/oxide')
     oxidePath = pathToFileURL(oxidePath).href
 
-    const oxide: typeof import('@tailwindcss/oxide') = await import(oxidePath)
+    const oxide: typeof import('@tailwindcss/oxide') = await import(oxidePath).then(
+      (o) => o.default || o,
+    )
```
thecrypticace, line 533: I don't think this is necessary? Did you see a reason to use `.default` here?

packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -500,7 +504,12 @@ async function* contentSelectorsFromCssConfig(entry: ConfigEntry): AsyncIterable
       }
     } else if (item.kind === 'auto' && !auto) {
       auto = true
-      for await (let pattern of detectContentFiles(entry.packageRoot)) {
+      let root = entry.entries[0]
```
thecrypticace, line 507: Took me a second to realize this was pointing to the CSS file entry (but it makes sense). I think a comment clarifying it would be useful.

packages/tailwindcss-language-server/src/resolve-css-imports.ts
```diff
@@ -15,8 +18,129 @@ const resolveImports = postcss([
...
+      source: fixRelativePath,
+      plugin: fixRelativePath,
+    },
+  }
+}
+
+// Inlined version of `normalize-path` <https://github.com/jonschlinkert/normalize-path>
+// Copyright (c) 2014-2018, Jon Schlinkert.
+// Released under the MIT License.
+function normalizePathBase(path: string, stripTrailing?: boolean) {
```
thecrypticace, line 103: I'm like 99% sure this already exists in this codebase

packages/tailwindcss-language-server/src/oxide.ts
```diff
@@ -0,0 +1,112 @@
...
+    resultV2 = oxide.scanDir({
+      base: options.basePath,
+      sources: options.sources,
+    })
+  } catch {
+    resultV1 = oxide.scanDir({
+      base: options.basePath,
+      globs: true,
+    })
+  }
```
philipp-spiess, line 97: This raises the questions:

- Should we properly version this API? 
- When can we deprecate support for an alpha release version? Should we even support older alpha releases? I worry that if we don't define a supported range right now, this will creep and become an unnecessary burden over time. 

Maybe we can say all alphas are supported for now but once we release something more stable, we'll drop support for it and will expect our users to upgrade? WDYT?

philipp-spiess: Since we have the load path, maybe we could read the `package.json` file to see what version we are working with 🤔

thecrypticace: 1. Should we properly version this API?
Well… we can't do that exactly because it's already released so…

2. When can we deprecate support for an alpha release version?

Good question — I think maybe once we're closer to a more stable API + a month or so we can drop stuff then.

> Since we have the load path, maybe we could read the package.json file to see what version we are working with 

Yeah that is an option 🤔

thecrypticace (approved):

PR #1031, opened 2024-08-09
Handle an async design system

thecrypticace:
In v4 we've made compile async as well as `__unstable_loadDesignSystem` and we need to handle that in intellisense.

### DIFF
--- a/packages/tailwindcss-language-server/src/util/v4/design-system.ts
+++ b/packages/tailwindcss-language-server/src/util/v4/design-system.ts
@@ -37,7 +37,11 @@ export async function loadDesignSystem(
   let resolved = await resolveCssImports().process(css, { from: filepath })
 
   // Step 3: Take the resolved CSS and pass it to v4's `loadDesignSystem`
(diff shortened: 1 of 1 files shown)


packages/tailwindcss-language-server/src/util/v4/design-system.ts
```diff
@@ -37,7 +37,11 @@ export async function loadDesignSystem(
   let resolved = await resolveCssImports().process(css, { from: filepath })
 
   // Step 3: Take the resolved CSS and pass it to v4's `loadDesignSystem`
-  let design = tailwindcss.__unstable__loadDesignSystem(resolved.css) as DesignSystem
+  let design: DesignSystem = await tailwindcss.__unstable__loadDesignSystem(resolved.css, {
+    loadPlugin() {
+      return () => {}
```
thecrypticace, line 42: We're intentionally not handling `@plugin` here — this PR is purely to make sure alpha 19 doesn't break in intellisense. That support will be done in a separate PR.

RobinMalfait (approved):

PR #1043, opened 2024-08-28
Handle postcss throwing when parsing more complex files

philipp-spiess:


lrungg:
for people with that issue: Just switch to the pre-release version of the extension, worked for me.

philipp-spiess:
@lrungg The fix or this was already released as version 0.12.8 so it should be fine to update to the latest stable version. 🙂

medz:
Oh my god! I never thought it was a problem with Tailwind SLP! I changed four or five editors because of this problem, and I always thought there was a problem with my project! In the past three days, I removed scss files every day to try to solve this problem! 😭

PR #1082, opened 2024-11-05
Lookup variables in the CSS theme

thecrypticace:
wip


packages/tailwindcss-language-service/src/util/css-vars.test.ts
```diff
@@ -1,26 +1,53 @@
...
 
   expect(
     replaceCssVarsWithFallbacks(
+      state,
       'color-mix(in srgb, var(--color-primary, oklch(0 0 0 / 2.5)), var(--color-secondary, oklch(0 0 0 / 2.5)), 50%)',
     ),
   ).toBe('color-mix(in srgb,  oklch(0 0 0 / 2.5),  oklch(0 0 0 / 2.5), 50%)')
+
+  // Fallbacks take precedence over theme values
+  expect(replaceCssVarsWithFallbacks(state, 'var(--foo, red)')).toBe(' red')
```
philipp-spiess, line 46: I understand that this is unlikely to be generated by TW's core utilities but I’m not sure I understand the concept behind it even. Shouldn't the theme var always get precedence just like it would in the browser? 🤔

philipp-spiess (commented):
LGTM but agree this needs to be tested as you pointed out 👍

packages/tailwindcss-language-service/src/util/css-vars.test.ts
thecrypticace: Oh yeah that's a good point. I was basically making a change that lines up with how intellisense _reads things today_ but you're right the value from the theme taking precedence makes sense. I'll swap the behavior here.

philipp-spiess (approved):

PR #1084, opened 2024-11-08
Support detection of deprecated classes

thecrypticace:
IntelliSense counterpart of https://github.com/tailwindlabs/tailwindcss/pull/14915


thecrypticace:
I'm gonna close this as the mechanism for detecting deprecated classes might actually change or maybe be more complicated in the case of a class being based on a deprecated theme key.

I'll likely be able to re-use a chunk of this PR once we get ready to implement this again.

PR #1087, opened 2024-11-11
Auto-switch CSS files to `tailwindcss` language in valid projects

thecrypticace:
This is an alternative design to part of #947. In that PR we'd be setting a default for *all* CSS files whether they're in a Tailwind CSS project or not. Instead we:

1. Detect that an opened CSS file is in a Tailwind CSS project and switch the language to `tailwindcss`. If you manually set it back to the CSS language we won't switch it back to `tailwindcss` again until you reload VSCode.
2. Switch any documents after the language server has found a project. This ensures that when starting a new project and you've created a CSS file typing `@import "tailwindcss"` and saving it will then switch the language to `tailwindcss`.


thecrypticace:
@philipp-spiess thoughts on this design?

philipp-spiess (approved):

PR #1092, opened 2024-11-18
Show theme values in hovers and completions

### LINKED ISSUE #1089: Show pixel values for v4 spacing
In Tailwind v4, [the spacing scale is now tweakable](https://x.com/adamwathan/status/1855592067029737512) using a `--spacing` variable. 

image

This is cool but a downside is the intellisense no longer shows the pixel values:

image

instead of:

image

It would be nice if the extension could calculate these.

v1.94.2

For example: v0.12.13

For example: v4.0.0-alpha.33

pnpgm

macOS

thecrypticace:
Fixes #1089

In v4 when we switched to using just `var(--some-theme-key)` in declarations instead of including fallbacks this had the side-effect of hiding pixel and color equivalents. This PR brings those back.


RobinMalfait (approved):
Love it. Excited to get a new beta release out and drop those undefined values.

thecrypticace:
Fantastic. I'm going to make some changes to the completion snippets so they're more usable.

packages/tailwindcss-language-server/tests/completions/completions.test.js
```diff
@@ -586,8 +586,12 @@ withFixture('v4/basic', (c) => {
 
     expect(resolved).toEqual({
       ...item,
-      detail:
-        'font-size: var(--text-sm); line-height: var(--tw-leading, var(--text-sm--line-height)); letter-spacing: undefined; font-weight: undefined;',
+      detail: [
+        'font-size: 0.875rem /* 8.75px */;',
+        'line-height: ≈ 1.4286;',
```
RobinMalfait, line 591: I wonder if this symbol here is going to be confusing or not (because it's not in a comment).

But I do like it though haha

RobinMalfait (approved):

packages/tailwindcss-language-server/tests/completions/completions.test.js
thecrypticace: I went back-and-forth on this one for this very reason. I think keeping it is OK. @reinink what do you think?

reinink: Hmm, I kind of think we can just leave the symbol out and just show the approximate value. Alternatively, could we just show the value here that's in the theme like `calc(1.25 / 0.875)`? To me that's the most readable version.

PR #1122, opened 2025-01-14
fix(language-service): allow native esm to import package

AviVahl:
with this patch, `@tailwindcss/language-service` can be imported/used in Node.js in native esm mode.

adjusted "semver" imports to include an extension, as that package has no "exports".

also upgrade semver to latest stable (7.6.3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "esbuild": "^0.20.2",
     "minimist": "^1.2.8",
     "prettier": "^3.2.5",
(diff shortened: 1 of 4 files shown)


thecrypticace (approved):

thecrypticace:
Thanks! This fix will go out once I do another release.

Rashelle:
Was looking for exactly that! @AviVahl  👏

PR #1131, opened 2025-01-22
fix(vscode-tailwindcss): fix activation events

gperdomor:
Currently the extension is activated always, even in not js/ts workspaces...

With this PR, the extension is activated when  a tailwind config file is present in the workspace or when `tailwindcss` is inside `node_modules/bin` folder

Fix https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1132

### DIFF
--- a/packages/vscode-tailwindcss/package.json
+++ b/packages/vscode-tailwindcss/package.json
@@ -35,7 +35,8 @@
   },
   "icon": "media/icon.png",
   "activationEvents": [
-    "onStartupFinished"
+    "workspaceContains:**/*{tailwind}*.config*.{js,cjs,mjs,ts,cts,mts}",
+    "workspaceContains:node_modules/.bin/tailwindcss"
   ],
   "main": "dist/extension.js",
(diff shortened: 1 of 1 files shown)


gperdomor:
@thecrypticace please can you review this please? 🙏🏻

thecrypticace:
Hey, so this isn't correct for this extension unfortunately. There's a few reasons:
- Tailwind v3 projects using Yarn PnP may not have a node_modules directory *at all*
- Tailwind v4 projects may not have a tailwind config file in js/mjs/ts files (in fact, most will not)
- Tailwind v4 projects will not have a CLI binary when using PostCSS or Vite unless the `@tailwindcss/cli` package is also installed.

Is there a problem you're seeing caused by it activating in non-Tailwind projects? It shouldn't launch the language server unless there's a possibility of it detecting a Tailwind project. In which case the language server will do some additional work and then bail if it's not detected.

gperdomor:
Ok I understand... Thanks for the explanation

PR #1142, opened 2025-01-26
feat: annotation support to highlight valid's className

hyoban:
ScreenShot 2025-01-26 16 48 20

This is very helpful for us to figure out whether the className is correct. I may not do well in some places, please point it out; I am happy to modify and improve it.

### DIFF
--- a/packages/tailwindcss-language-server/src/config.ts
+++ b/packages/tailwindcss-language-server/src/config.ts
@@ -16,6 +16,7 @@ function getDefaultSettings(): Settings {
       classAttributes: ['class', 'className', 'ngClass', 'class:list'],
       codeActions: true,
       hovers: true,
+      annotations: false,
       suggestions: true,
       validate: true,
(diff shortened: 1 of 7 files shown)


thecrypticace:
Hah, fun fact I actually had a "debug" version of this idea at one point to help me visualize individual classes vs class lists when using custom regexes while I was debugging some issues.

I **really** like this idea in general but I'll need to think on the implementation some as I think there are two main ways to approach this:
- Use it to analyze what "spans" of text are being matched by custom class regexes
- Use it to analyze what individual classes are being matched by Tailwind CSS itself / Oxide.

I think both of these approaches could use useful and could also be nice power-user features to have.

Also wanted to let you know it may be a bit before I get back to this PR but in general — very much like the idea. I have some ideas for changing how documents are handled in IntelliSense to simplify / unify the implementations of several features. I think this could be a good feature to add after that refactor.

PR #1159, opened 2025-01-31
Skip over preprocessor files when looking for v4 configs

thecrypticace:
Tailwind CSS v4.0 is more like a preprocessor and [should not be used with Sass, Less, or Stylus](https://tailwindcss.com/docs/compatibility#sass-less-and-stylus). We don't want to consider these files to be potential CSS configuration files for a v4 project.

### DIFF
--- a/packages/tailwindcss-language-server/src/project-locator.ts
+++ b/packages/tailwindcss-language-server/src/project-locator.ts
@@ -751,5 +751,5 @@ class FileEntry {
 function requiresPreprocessor(filepath: string) {
   let ext = path.extname(filepath)
 
-  return ext === '.scss' || ext === '.sass' || ext === '.less' || ext === '.styl'
+  return ext === '.scss' || ext === '.sass' || ext === '.less' || ext === '.styl' || ext === '.pcss'
 }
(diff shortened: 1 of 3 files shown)


thecrypticace:
This technically is based on another PR that's not quite ready yet but I wanted to at least get eyes on these changes. Plan to merge this in next week.

packages/tailwindcss-language-server/src/project-locator.ts
```diff
@@ -733,3 +747,9 @@ class FileEntry {
     )
   }
 }
+
+function requiresPreprocessor(filepath: string) {
+  let ext = path.extname(filepath)
+
+  return ext === '.scss' || ext === '.sass' || ext === '.less' || ext === '.styl'
```
RobinMalfait, line 754: Do we also want to include `.pcss` or is that not an issue?

RobinMalfait (approved):
Makes sense, just a question

packages/tailwindcss-language-server/src/project-locator.ts
thecrypticace: My thinking was that `.pcss` files (some extension for postcss) the files should still be valid CSS. Buuuut… yeah maybe we should include it anyway. I can relax the check later if need be.

philipp-spiess (approved):

Issue #926, opened 2024-03-17
Hovers and completions not showing in a v4 project


thecrypticace:
From [Discord](https://discord.com/channels/486935104384532500/1217407069053780028/1218820631575924766):

```
Hi. I discovered what the cause was but I'm not sure it's really relevant or useful. So I'm not sure if I should still file an Issue for it.

Basically, in my project I have `src/styles/styles.css` with the new v4 syntax. Then, for reference, I copied a tailwind v3 base css file into `src/styles/_global.css`

`_global.css` is completely unused in the project.

When you reload vs code, Tailwind intellisense and hovers break in the way i described previously.

So if the v3 .css file comes before the v4 css file, it breaks. but if it comes after, nothing happens.

Example that works
- `styles.css`
- `unused-v3.css` <- comes after the v4 file alphabetically
https://github.com/Brian-Pob/tailwind-v4-vite/tree/c2520ecd266e6de8a14df127f77ee55ad6322997/src/styles

Example that breaks
- `my-unused-v3.css` <- comes before the v4 file alphabetically
- `styles.css`
https://github.com/Brian-Pob/tailwind-v4-vite/tree/54c23c939d04d247844f02cc628a83ca66c71459/src/styles

Astro                    v4.5.4
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          bun
Output                   static
Adapter                  none
Integrations             @astrojs/svelte

tailwindcss@4.0.0-alpha.9
@tailwindcss/vite@4.0.0-alpha.9

VS Code Version: 1.87.2
```

Brian-Pob:
Hello. I'm the one who brought up this issue on Discord. Thank you for creating the issue for me.

PR #1164, opened 2025-02-04
Don't create v4 projects for CSS files that don't look like v4 configs

thecrypticace:
Fies #926

### DIFF
--- a/packages/tailwindcss-language-server/src/version-guesser.ts
+++ b/packages/tailwindcss-language-server/src/version-guesser.ts
@@ -1,0 +1,48 @@
+export type TailwindVersion = '3' | '4'
+
+/**
+ * Determine the likely Tailwind version used by the given file
+ *
+ * This returns an array of possible versions, as a file could contain
+ * features that make determining the version ambiguous.
+ *
+ * The order *does* matter, as the first item is the most likely version.
+ */
+export function guessTailwindVersion(content: string): TailwindVersion[] {
+  if (!content) return []
+
+  // It's likely this is a v4 file if it has a v4 import:
+  // - `@import "tailwindcss"`
+  // - `@import "tailwindcss/theme"
+  // - etc…
+  let HAS_V4_IMPORT = /@import\s*['"]tailwindcss(?:\/[^'"]+)?['"]/
+  if (HAS_V4_IMPORT.test(content)) return ['4']
+
+  // It's likely this is a v4 file if it has a v4-specific feature:
+  // - @theme
+  // - @plugin
+  // - @utility
+  // - @variant
+  // - @custom-variant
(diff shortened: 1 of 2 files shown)


thecrypticace:
@RobinMalfait @philipp-spiess what do you think about this approach?

packages/tailwindcss-language-server/src/version-guesser.ts
```diff
@@ -0,0 +1,48 @@
...
+  let HAS_V4_IMPORT = /@import\s*['"]tailwindcss(?:\/[^'"]+)?['"]/
+  if (HAS_V4_IMPORT.test(content)) return ['4']
+
+  // It's likely this is a v4 file if it has a v4-specific feature:
+  // - @theme
+  // - @plugin
+  // - @utility
+  // - @variant
+  // - @custom-variant
+  let HAS_V4_DIRECTIVE = /@(theme|plugin|utility|custom-variant|variant)\s*[^;{]+[;{]/
```
RobinMalfait, line 27: I think we can add `@reference` to this list as well

packages/tailwindcss-language-server/src/version-guesser.ts
```diff
@@ -0,0 +1,48 @@
...
+  let HAS_TAILWIND = /@tailwind\s*[^;]+;/
+  if (HAS_TAILWIND.test(content)) return ['4', '3']
+
+  // If the file contains other `@apply` or `@config` it might be either
+  let HAS_COMMON_DIRECTIVE = /@(config|apply)\s*[^;{]+[;{]/
+  if (HAS_COMMON_DIRECTIVE.test(content)) return ['4', '3']
+
+  // If it's got imports at all it could be either
+  let HAS_IMPORT = /@import\s*['"]/
+  if (HAS_IMPORT.test(content)) return ['4', '3']
```
RobinMalfait, line 44: Some other heuristics for v4 would be detecting `--theme()`, `--alpha()` and `--spacing()`.

These will/should probably be used in combination with an `@import` and/or a `@reference`. In case of an `@import` currently we are saying that it could be either `4` or `3`, but if you see these functions listed above, we know for sure it's a v4 project. 

Thoughts?

packages/tailwindcss-language-server/src/version-guesser.ts
thecrypticace: OH… yep yep yep thanks good call

packages/tailwindcss-language-server/src/version-guesser.ts
thecrypticace: Yeah that's a good idea. I'll add that heuristic to the list.

thecrypticace: Done

RobinMalfait (approved):

philipp-spiess (approved):

PR #1165, opened 2025-02-04
Improve CSS completions

### LINKED ISSUE #1146: [v4] `@utility` directive's content css doesn't have intellisense for css properties

4.0.0

**What build tool (or framework if it abstracts the build tool) are you using?**

vite 6.0.11

**What version of Node.js are you using?**

20.13.1

**What browser are you using?**

Chrome

macOS

**Describe your issue**

as stated in the title.
when using the new `@utility` directive, the inner css code doesn't have any intellisense. its like we are not using any css code.
(the code itself is working just fine, its only an intellisense issue)

in the image below, we can even see that the coloring (plain white) indicates it
Image

thecrypticace:
- [x] Support property and variable completions inside `@utility` (behave like a style rule)
- [x] Support style-rule like completions inside `@custom-variant`
- [x] Support style-rule like completions inside `@variant`
- [x] Make sure `@slot` isn't considered an unknown at-rule

Fixes #1146

### DIFF
--- a/packages/tailwindcss-language-server/src/language/cssServer.ts
+++ b/packages/tailwindcss-language-server/src/language/cssServer.ts
@@ -409,7 +409,7 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
     .filter((diagnostic) => {
       if (
         diagnostic.code === 'unknownAtRules' &&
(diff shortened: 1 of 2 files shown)


philipp-spiess (commented):
Add `@variant` and we gucci

philipp-spiess (approved):

PR #1166, opened 2025-02-04
Fix equivalent calculation when using prefixes in v4

### LINKED ISSUE #1160: v4: Prefixed theme values are not used to calculate pixel equivalents

v1.96.4

v0.14.2

v4.0.3

bun

Windows

**Tailwind config**

```css
@import 'tailwindcss' prefix(tw);
```

**Describe your issue**

When hovering over a prefixed tailwind 4 class the Tailwind Intellisense hint does not show the correct hint:

Image

If tailwind config does not use a prefix;
```css
@import 'tailwindcss';
``` 
the Tailwind Intellisense hint works correctly:

Image

thecrypticace:
Fixes #1160

We were passing the CSS variable straight through when resolving the theme values but this isn't accurate when a prefix is used. The theme variables *do not* have prefixes but the emitted CSS variables _do_.

We have to strip the prefix before doing the lookup.


philipp-spiess (approved):
Works like a charm

PR #1168, opened 2025-02-04
Perform CSS variable resolution recursively

### LINKED ISSUE #1102: v4 swatches not displayed when referring variables
Cursor, and not sure where to see this

v0.13.59 (pre-release)

v4.0.0-alpha.34

npm

macOS

**Describe your issue**

Color swatches are not displaying when referring to variables from the main config file. It worked fine with Tailwind v3.

main.css
```css
    --color-test-500: #ff0000; 
    --color-primary-500: var(--color-indigo-50);
```

index.html
```html
<div class="bg-text-500">Swatch displays</div>
<div class="bg-primary-500">Swatch doesn't displays</div>
```

The auto-suggest works fine however, it's just the color swatch

thecrypticace:
Fixes #1102

### DIFF
--- a/packages/vscode-tailwindcss/CHANGELOG.md
+++ b/packages/vscode-tailwindcss/CHANGELOG.md
@@ -12,6 +12,9 @@
 - Make sure `@slot` isn't considered an unknown at-rule ([#1165](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1165))
 - Fix equivalent calculation when using prefixes in v4 ([#1166](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1166))
 - Fix use of `tailwindCSS.experimental.configFile` option when using the bundled version of v4 ([#1167](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1167))
+- Recursively resolve values from the theme ([#1168](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1168))
+- Handle theme keys containing escaped commas ([#1168](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1168))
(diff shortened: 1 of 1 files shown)


packages/tailwindcss-language-service/src/util/rewriting/replacements.ts
```diff
@@ -58,9 +77,20 @@ export function replaceCssVars(str: string, replace: CssVarReplacer): string {
...
+        // track of which variables we've already seen and where they were
+        // replaced to avoid infinite loops.
+        if (recursive) {
+          let key = `${i}:${replacement}`
+
+          if (!seen.has(key)) {
+            seen.add(key)
+            i -= 1
+          }
+        }
```
thecrypticace, line 92: @RobinMalfait or @philipp-spiess can y'all think of a way to break this? It has the potential for infinite recursion (due to circular theme key lookups) but I guarded against it here.

I think this should be sufficient but if you could give it some thought that would be much appreciated.

packages/tailwindcss-language-service/src/util/rewriting/replacements.ts
```diff
@@ -58,9 +77,20 @@ export function replaceCssVars(str: string, replace: CssVarReplacer): string {
           str = str.slice(0, i) + replacement + str.slice(j + 1)
```
philipp-spiess, line 77: Screenshot 2025-02-05 at 13 28 32

LGTM but we should handle `\` here 🙈

philipp-spiess (approved):

packages/tailwindcss-language-service/src/util/rewriting/replacements.ts
thecrypticace: Fixed that too

PR #1172, opened 2025-02-05
Allow hover over `--theme(…)` function in `@media` rule without spaces

philipp-spiess:
This PR fixes an issue with the regex detecting the `--theme(…)` function. If used within a media query with no spaces, it was not properly detecting word boundaries as characters like `<` etc. would be delimit properly.

## Test plan

Added a unit test to ensure this does not regress.

### DIFF
--- a/packages/tailwindcss-language-server/tests/hover/hover.test.js
+++ b/packages/tailwindcss-language-server/tests/hover/hover.test.js
@@ -272,6 +272,24 @@ withFixture('v4/basic', (c) => {
       end: { line: 2, character: 18 },
     },
   })
+
+  testHover('--theme() works inside @media queries', {
+    lang: 'tailwindcss',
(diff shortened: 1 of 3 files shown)


packages/tailwindcss-language-server/tests/hover/hover.test.js
```diff
@@ -272,6 +272,24 @@ withFixture('v4/basic', (c) => {
...
+      contents: {
+        kind: 'markdown',
+        value: ['```plaintext', '80rem /* 1280px */', '```'].join('\n'),
+      },
+      range: {
+        start: { line: 0, character: 23 },
+        end: { line: 0, character: 38 },
+      },
+    },
+  })
```
thecrypticace, line 292: Can we add a test for `theme()` as well?

thecrypticace (approved):

PR #1212, opened 2025-02-16
Fix odd v4 behavior with custom @-rules

### LINKED ISSUE #1211: scuffed @custom-variant behavior
Registering a v4 custom-variant with a nested (pseudo-)selector causes "at-rule or selector expected". It compiles just fine.

```css
@custom-variant poi {
	&:hover,
	&:focus-visible {
		@slot;
	}
}
```

Image

The last 2 errors go away on removing the `@slot`s `;`. However, this is not a viable option, as Prettier adds it back.

---

Oddly enough, it seems to replace this with... whatever this mess is.

Image

---

Image

This happens because the Regex matches up to a `;`: `(\s+[^;]+)`. I'm not sure what the point of this is.
There is another matcher that uses { right below, but it never gets called, as the matcher with the ; already
replaces everything up to the slot semicolon.

Removing it seems to have done the trick. All tests pass with this change.

---

Image

... now this throws. Even if I restore the original extension. lovely.

The space was added by Prettier, but this also happens without it. Adding other spaces restores it to `[ *]`, so my PR allows a space before `*`.

It seems like it doesn't like *. Replacing it with _ in the virtual document works fine. All tests pass with this change.

---

I'd like to work on this. I'll follow up with a PR in a bit.

tizu69:
Fixes #1211

I'm not sure if the `.replace` I removed actually had a purpose - it seemed useless to me.
I can (probably) easily bring it back if it does have a purpose, gonna have to see how to fix it in that case hten.

Also adds a QOL feature of now autosuggesting `[*]` in `--value(`.

See the issue for way more details on why some things are the way they are.

### DIFF
--- a/packages/tailwindcss-language-server/src/language/cssServer.ts
+++ b/packages/tailwindcss-language-server/src/language/cssServer.ts
@@ -386,14 +386,6 @@ function createVirtualCssDocument(textDocument: TextDocument): TextDocument {
     )
     .replace(/(?<=\b(?:theme|config)\([^)]*)[.[\]]/g, '_')
 
-    .replace(/--value\(\[(\s?)\*\]\)/g, (_match, space) => {
-      // In this case, space represents a single character space.
-      // From what I can tell spaces aren't intended, so I limited this to a single matcher.
-      // This is because Prettier likes to add a space before the `*`.
-      // FIXME: This is probably unintentional behavior. Consider fixing.
-      return `--value([${space}_])`
-    })
-
   return TextDocument.create(
     textDocument.uri,
     textDocument.languageId,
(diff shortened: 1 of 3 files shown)


thecrypticace:
The .replace _is_ necessary but it's matching too much. I've fixed it up. Will get this merged in the am after some more testing.

tizu69:
> The .replace _is_ necessary but it's matching too much. I've fixed it up. Will get this merged in the am after some more testing.

out of curiosity, what it is used for? I couldn't find anything that that would catch (or is supposed to catch).

philipp-spiess:
@tizu69 One is for the shorthand syntax, e.g.:

```css
@custom-variant dark (& .dark);
```

and one is for the syntax the block syntax: 

```css
@custom-variant dark {
  &:where(.dark, .dark *) {
    @slot;
  }
}
```

You can see that one regex contains the semicolon and one an opening-brace 👍

tizu69:
ah, I see. I wasn't familiar with that first syntax, fair enough.

thecrypticace (approved):

PR #1215, opened 2025-02-17
Don't show syntax error for `*`s inside `--value` and `--modifier` functions

### LINKED ISSUE #1208: Invalid css-termexpected error?
Context: https://github.com/Wombosvideo/tailwindcss-animate-v4/issues/1

I got the following syntax error

Image

on code like ([source](https://github.com/Wombosvideo/tailwindcss-animate-v4/blob/4dae3becb75e7c33242fa84306aab1bea0889483/tailwindcss-animate.css#L96-L108))

Image

The same syntax appears in the same file ([code](https://github.com/Wombosvideo/tailwindcss-animate-v4/blob/4dae3becb75e7c33242fa84306aab1bea0889483/tailwindcss-animate.css#L118-L130)) but the plugin only reports error on the 4 lines above. And removing the whitespaces before the asterisk doesn't fix the error.

Sorry for my ignorance on the CSS syntax, is this a bug on the plugin or how should the code to be written in a correct way?

thecrypticace:
Fixes #1208

Needs infrastructure from #1216 and tests

### DIFF
--- a/packages/tailwindcss-language-server/src/language/rewriting.ts
+++ b/packages/tailwindcss-language-server/src/language/rewriting.ts
@@ -68,5 +68,10 @@ export function rewriteCss(css: string) {
 
   css = css.replace(/(?<=\b(?:theme|config)\([^)]*)[.[\]]/g, '_')
 
+  // Ignore `*` in in --value and --modifier functions
+  css = css.replace(/--(value|modifier)\((.*?)\)/g, (match) => {
+    return match.replace(/[*]/g, '_')
+  })
+
   return css
 }
--- a/packages/tailwindcss-language-server/src/language/rewriting.ts
+++ b/packages/tailwindcss-language-server/src/language/rewriting.ts
@@ -73,7 +73,7 @@ export function rewriteCss(css: string) {
     return match.replace(/[*]/g, '_')
   })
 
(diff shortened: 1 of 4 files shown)


packages/tailwindcss-language-server/src/language/rewriting.ts
```diff
@@ -68,5 +68,13 @@ export function rewriteCss(css: string) {
 
   css = css.replace(/(?<=\b(?:theme|config)\([^)]*)[.[\]]/g, '_')
 
+  // Ignore `*` in in --value and --modifier functions
+  css = css.replace(/--(value|modifier)\((.*?)\)/g, (match) => {
+    return match.replace(/[*]/g, '_')
+  })
+
+  // Replace `--some-var-*` with `--some-var_`
```
RobinMalfait, line 76:
```suggestion
  // Replace `--some-var-*` with `--some-var-_`
```

RobinMalfait (approved):

packages/tailwindcss-language-server/src/language/rewriting.ts
thecrypticace: Nope, the comment is right.

PR #1216, opened 2025-02-17
Add Tailwind CSS language mode tests

thecrypticace:
This PR does a small bit of refactoring on the bundled CSS language server and adds tests for it to ensure we don't break stuff


packages/tailwindcss-language-server/src/language/css-server.ts
```diff
@@ -0,0 +1,379 @@
...
+    })
+    documents.onDidClose(({ document }) => {
+      stylesheets.onDocumentRemoved(document)
+    })
+    connection.onShutdown(() => {
+      stylesheets.dispose()
+    })
+
+    connection.onInitialize((params: InitializeParams) => {
+      workspaceFolders = (<any>params).workspaceFolders
```
RobinMalfait, line 56: I don't think I've ever used a cast like that in TypeScript ever haha. Think I always do this:
```suggestion
      workspaceFolders = (params as any).workspaceFolders
```

Just a comment, no change required though.

RobinMalfait (approved):
Not 100% sure how to test it properly apart from running it locally in vscode and it seems to work as expected.

Also looks like most of it is moving things around + cleanup.

Additional tests are also nice and no older tests break so that also a win.

PR #1226, opened 2025-02-20
Rebuild testing setup

thecrypticace:
This PR basically rebuilds the testing infra with code in place to keep existing tests working.

This was done so we can add tests for the CSS language server in a followup PR.


RobinMalfait (approved):
I just ran this locally on a project and all still seems to work.

image
image

PR #1227, opened 2025-02-20
Show `@theme` in symbol list

thecrypticace:
Right now `@theme` is treated as an opaque at-rule by the CSS language server and it doesn't show up in symbol navigation. This PR addresses this by performing the same type of replacement we do for `@utility`, `@variant`, and `@custom-variant`.

### DIFF
--- a/packages/tailwindcss-language-server/tests/css/css-server.test.ts
+++ b/packages/tailwindcss-language-server/tests/css/css-server.test.ts
@@ -203,6 +203,46 @@ defineTest({
 })
 
 defineTest({
+  name: '@theme',
+  prepare: async ({ root }) => ({
+    client: await createClient({
+      server: 'css',
+      root,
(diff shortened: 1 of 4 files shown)


packages/tailwindcss-language-server/tests/css/css-server.test.ts
```diff
@@ -202,6 +202,46 @@ defineTest({
...
+        @theme {
+          --color-primary: #333;
+        }
+      `,
+    })
+
+    // No errors
+    expect(await doc.diagnostics()).toEqual([])
+
+    // Symbols show up for @custom-variant
```
RobinMalfait, line 228:
```suggestion
    // Symbols show up for @theme
```

RobinMalfait (approved):
Nice, small copy/pate issue

PR #1237, opened 2025-02-26
Fix syntax error when resetting multi-word theme key namespaces

### LINKED ISSUE #1236: Regression in v0.14.7: Font weight reset of defaults in `@theme` shows errors
With the update to `v0.14.7`, the following line, and all lines below it, are showing errors.

```css
--font-weight-*: initial;
```

Removing the above line also resolves all errors in subsequent lines below it. The CSS itself is valid, and works correctly with Tailwind. This is a problem with the VS Code extension.

## Screeshots:

Image

Image

Image

## Regression

This worked fine until `v0.14.6` where there were no errors. See screenshot when the older version is installed:

Image

Current workaround: Downgrade to `v0.14.6` and disable auto-update until this is resolved.

## Additional Information

This doesn't happen with other resets above the font weight reset. I have several others above it, like:

```css
--color-*: initial;
--breakpoint-*: initial;
--radius-*: initial;
--shadow-*: initial;
--font-*: initial;
--font-weight-*: initial; /* <-- Errors in v0.14.7 start here. */
```

If I move the font weight reset to be the first line in the `@theme` block, then everything below it gets errors, including all those other resets.

It looks like a parsing problem when there are two words instead of just one. For example, if I change the color reset to `--color-test-*: initial;` it starts having the same error too.

thecrypticace:
Fixes #1236

This PR fixes an issue where we'd show a syntax error for multi-word namespace resets like `--font-weight-*: initial;`:

```css
@theme {
  --font-*: initial; /* this one is fine */
  --font-weight-*: initial; /* this one shows a syntax error */
}
```

Screenshot 2025-02-26 at 12 16 38

Now both work as expected:

Screenshot 2025-02-26 at 12 17 09

### DIFF
--- a/packages/tailwindcss-language-server/src/language/rewriting.ts
+++ b/packages/tailwindcss-language-server/src/language/rewriting.ts
@@ -74,7 +74,7 @@ export function rewriteCss(css: string) {
   })
 
   // Replace `--some-var-*` with `--some-var-_`
-  css = css.replace(/--([a-zA-Z0-9]+)-[*]/g, '--$1_')
+  css = css.replace(/--([a-zA-Z0-9-]+)-[*]/g, '--$1-_')
 
   return css
 }
--- a/packages/tailwindcss-language-server/src/language/rewriting.test.ts
+++ b/packages/tailwindcss-language-server/src/language/rewriting.test.ts
@@ -69,20 +69,28 @@ test('@theme', () => {
   let input = [
     //
     '@theme {',
-    '  color: red;',
+    '  --color: red;',
+    '  --font-*: initial;',
+    '  --font-weight-*: initial;',
     '}',
     '@theme inline reference static default {',
-    '  color: red;',
+    '  --color: red;',
+    '  --font-*: initial;',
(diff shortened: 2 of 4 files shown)


packages/tailwindcss-language-server/src/language/rewriting.ts
```diff
@@ -74,7 +74,7 @@ export function rewriteCss(css: string) {
   })
 
   // Replace `--some-var-*` with `--some-var-_`
-  css = css.replace(/--([a-zA-Z0-9]+)-[*]/g, '--$1_')
+  css = css.replace(/--([a-zA-Z0-9-]+)-[*]/g, '--$1-_')
```
philipp-spiess, line 77: hehe

philipp-spiess (approved):

packages/tailwindcss-language-server/src/language/rewriting.ts
thecrypticace: right?! lol

Issue #1224, opened 2025-02-20
Tailwind intellisense does not work natively with tailwindcss-ruby when using Tailwind V4.X


ryancbahan:
I'm upgrading a few apps that use [tailwindcss-rails](https://github.com/rails/tailwindcss-rails?tab=readme-ov-file#upgrading-your-application-from-tailwind-v3-to-v4) to Tailwind v4. The tailwind rails gem uses [tailwindcss-ruby](https://github.com/flavorjones/tailwindcss-ruby), which vendors the Tailwind executable rather than getting it via npm or node_modules. 

This seems to be breaking tailwind intellisense, which appears to always be looking for the tailwind executable in node_modules. When I put a debugger in the extension, I can see output:

```
Error: Can't resolve 'tailwindcss' in '/Users/ryan/world/trees/root/src/areas/core/shopify/components/developer_dashboard/app/assets/tailwind'
    at finishWithoutResolve (/Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:2939:25)
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3023:26
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3073:13
    at eval (eval at create (/Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:1542:19), <anonymous>:15:1)
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3073:13
    at eval (eval at create (/Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:1542:19), <anonymous>:27:1)
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3878:52
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3073:13
    at eval (eval at create (/Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:1542:19), <anonymous>:15:1)
    at /Users/ryan/tailwindcss-intellisense/packages/vscode-tailwindcss/dist/server.js:3073:13 {
[...]

PR #1240, opened 2025-02-27
v4: Support loading bundled versions of some first-party plugins

thecrypticace:
see https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1224#issuecomment-2687637180

This adds support for loading these three plugins in v4 **only when using `@plugin`** if we can't find them:
- `@tailwindcss/typography`
- `@tailwindcss/forms`
- `@tailwindcss/aspect-ratio`

This coincides with behavior of the Standalone CLI where these are bundled instead of available as an NPM package.

I've additionally logs when trying to import one of these plugins inside a JS file. This does not work right now and may take some effort to support because it will at least require the use of Node's experimental loader's API and potentially additional work as well.

### DIFF
--- a/packages/tailwindcss-language-server/src/util/v4/design-system.ts
+++ b/packages/tailwindcss-language-server/src/util/v4/design-system.ts
@@ -74,7 +74,9 @@ function createLoader<T>({
         if (match) {
           let [_, importee] = match
           if (importee in plugins) {
-            console.log('Loading plugins inside configs or plugins is not currently supported')
+            console.log(
+              `[error] Cannot load '${id}' plugins inside configs or plugins is not currently supported`,
+            )
           }
         }
       }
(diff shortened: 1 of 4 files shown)


packages/tailwindcss-language-server/src/util/v4/design-system.ts
```diff
@@ -58,6 +59,28 @@ function createLoader<T>({
 
       return await jiti.import(url.href, { default: true })
     } catch (err) {
+      // If the request was to load a first-party plugin and we can't resolve it
+      // locally, then fall back to the built-in plugins that we know about.
+      if (resourceType === 'plugin' && id in plugins) {
+        console.log('Loading bundled plugin for: ', id)
```
philipp-spiess, line 65: Did you keep these logs on purpose?

philipp-spiess (approved):

packages/tailwindcss-language-server/src/util/v4/design-system.ts
thecrypticace: Yep. We send these to the "Tailwind CSS IntelliSense" output panel in VSCode. It's useful for triaging issues.

PR #1252, opened 2025-03-05
Don’t throw when client is missing the `textDocument` capability

### LINKED ISSUE #1251: Language Server throws when `textDocument` client capability is missing

N/A

0.14.8

N/A

npm

macOS

**Tailwind config**

N/A

**VS Code settings**
N/A

**Reproduction**

To reproduce, run the `lspx` shell against any version of the language server: 

```
deno run -A jsr:@frontside/lspx --interactive --lsp "tailwindcss-language-server --stdio"
```

Then, when prompted enter the command `initialize({"capabilities": {}})` into the prompt:

```
LSP> initialize({"capabilities": {}})
```

**Describe your issue**

The tailwind language server throws an error when there are no `textDocument` capabilities on the client to which it is attached. 

```
LSP> initialize({"capabilities": {}})
Sending request initialize with params [{"capabilities":{}}]...
Error: Request initialize failed with message: Cannot read properties of undefined (reading 'hover')
```

The spec indicates that textDocument is an optional property https://github.com/microsoft/vscode-languageserver-node/blob/df05883f34b39255d40d68cef55caf2e93cff35f/protocol/src/common/protocol.ts#L1098

While it is very rare that this should happen, it is still a possibility, and the language server should handle it gracefully.

thecrypticace:
Fixes #1251

This needs more research before merging because we probably need to skip registration of providers. And possibly not even attempt to initialize projects in this case? Not entirely sure.

- [ ] Needs tests

### DIFF
--- a/packages/tailwindcss-language-server/src/tw.ts
+++ b/packages/tailwindcss-language-server/src/tw.ts
@@ -1024,10 +1024,10 @@ export class TW {
 
 function supportsDynamicRegistration(params: InitializeParams): boolean {
   return (
-    params.capabilities.textDocument.hover?.dynamicRegistration &&
-    params.capabilities.textDocument.colorProvider?.dynamicRegistration &&
-    params.capabilities.textDocument.codeAction?.dynamicRegistration &&
-    params.capabilities.textDocument.completion?.dynamicRegistration &&
-    params.capabilities.textDocument.documentLink?.dynamicRegistration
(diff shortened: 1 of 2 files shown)


thecrypticace:
I'm gonna skip the automated tests for now. I can't find *anything* reasonable in the language server spec about how things should act in this case so I'm just gonna wait for people to open an issue if that ever happens.

RobinMalfait (approved):

philipp-spiess (approved):

PR #1267, opened 2025-03-16
V4: Update README.md

junepil:
With the update to Tailwind CSS v4, [JavaScript config files](https://tailwindcss.com/docs/upgrade-guide#using-a-javascript-config-file) are still supported for backward compatibility but are now redundant. Because of this, I believe the README.md should be updated to avoid confusion, especially when debugging issues like IntelliSense not working.

Additionally, the following link currently points to the [theme documentation](https://tailwindcss.com/docs/theme), which seems incorrect. It should either be updated to the [v3 configuration guide](https://v3.tailwindcss.com/docs/configuration#creating-your-configuration-file) if relevant or removed altogether.

Thanks for the great project! 🤟

### DIFF
--- a/packages/vscode-tailwindcss/README.md
+++ b/packages/vscode-tailwindcss/README.md
@@ -6,7 +6,7 @@ Tailwind CSS IntelliSense enhances the Tailwind development experience by provid
 
 **[Install via the Visual Studio Code Marketplace →](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss)**
 
-In order for the extension to activate you must have [`tailwindcss` installed](https://tailwindcss.com/docs/installation) and a [Tailwind config file](https://tailwindcss.com/docs/installation#create-your-configuration-file) named `tailwind.config.{js,cjs,mjs,ts,cts,mts}` in your workspace.
+In order for the extension to activate you must have [`tailwindcss` installed](https://tailwindcss.com/docs/installation).
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


nyatinte:
Hi @junepil,

Thanks a lot for this pull request! Updating the README to better reflect Tailwind CSS v4's configuration is really valuable and helps prevent confusion for users.

Actually, I've also experienced some issues related to v4 configuration detection myself, and looking at this PR helped me understand the situation better.

I agree that the previous wording focusing solely on the `tailwind.config.{...}` file as a strict requirement might be misleading for v4 projects, especially those using the CSS-first approach. Removing that outdated emphasis is a great step.

Although I'm not a formal reviewer, from a user's perspective, I believe adding some further explanation in the README would be beneficial, particularly regarding Tailwind CSS v4 configuration detection.

Based on the code (e.g., in [analyze.ts](https://github.com/tailwindlabs/tailwindcss-intellisense/blob/main/packages/vscode-tailwindcss/src/analyze.ts)), the extension appears to detect configuration not only from the traditional JavaScript config file but also from CSS files containing v4 directives like `@import` and `@theme`.

To clarify this for users, especially those starting with v4's CSS-first approach, I suggest updating the activation criteria description to cover both methods. Perhaps something like this:

```md

"In order for the extension to activate, you must have `tailwindcss` installed in your workspace. The extension will then attempt to detect your Tailwind CSS configuration, which can be located in one of the following:

* A JavaScript config file (`tailwind.config.{js,cjs,mjs,ts,cts,mts}`), typically used in **v3 and earlier**.
* Or, for Tailwind CSS **v4** projects, configuration defined directly within your main CSS file using directives like `@import "tailwindcss";` and `@theme { ... }`.

```

This explicitly acknowledges the v4 CSS-first method and could help users understand how the extension identifies the configuration source.

What are your thoughts on incorporating this kind of detail into the README?

junepil:
Thank's for the suggestion @nyatinte , I also believe that your version is definitely more user friendly and leads to correct understanding of the process using this extension.

thecrypticace:
o/ For some reason I can't push to this PR even tho it says I should. I've copied the branch over to a new PR and started making changes there so you're still credited. Thanks!

See #1364

PR #1269, opened 2025-03-18
Bump `@parcel/watcher`

thecrypticace:
This is just a general version bump. I thought an issue was parcel watcher related but turned out not to be. Saw we were **very** behind so I wanted to make sure we were up to date.

### DIFF
--- a/packages/tailwindcss-language-server/package.json
+++ b/packages/tailwindcss-language-server/package.json
@@ -34,7 +34,15 @@
     "access": "public"
   },
   "devDependencies": {
-    "@parcel/watcher": "2.0.3",
+    "@parcel/watcher": "2.5.1",
+    "@parcel/watcher-darwin-x64": "2.5.1",
+    "@parcel/watcher-darwin-arm64": "2.5.1",
+    "@parcel/watcher-win32-x64": "2.5.1",
+    "@parcel/watcher-win32-arm64": "2.5.1",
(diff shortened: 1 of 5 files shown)


packages/tailwindcss-language-server/src/testing/index.ts
```diff
@@ -1,13 +1,16 @@
...
 import * as path from 'node:path'
 import * as proc from 'node:child_process'
 import dedent from 'dedent'
 
-export interface TestUtils {
+export interface TestUtils<TestInput extends Record<string, any>> {
   /** The "cwd" for this test */
   root: string
+
+  /** The "cwd" for this test */
```
philipp-spiess, line 12: copy-paste error? I think `root` is already the `"cwd"` for this test

philipp-spiess (approved):
Mind including some context as to why this is necessary or if this is just a general version bump etc?

packages/tailwindcss-language-server/src/testing/index.ts
thecrypticace: I swore I fixed this lol thanks will update

thecrypticace:
> Mind including some context as to why this is necessary or if this is just a general version bump etc?

Done. Yeah, it's just a general version bump since we're suuuuuuper behind. It was born out of triaging an issue that I thought was parcel watcher related but it ended up that it wasn't.

PR #1274, opened 2025-03-19
v4: Show theme key completions in `var(…)`

### LINKED ISSUE #1209: [Feature]: Provide VS Code Autocompletion for Tailwind's Generated CSS Variables
So far, I've been amazed and very appreciative of Tailwind's abilities to provide autocomplete for its utility classes. And with the [right VS Code configuration](https://github.com/paolotiu/tailwind-intellisense-regex-list?tab=readme-ov-file#dom), you can accomplish even more!

I think the main limitation right now is autocompletion in the CSS files. According to the [documentation](https://tailwindcss.com/docs/upgrade-guide#using-the-theme-function), we should prefer `var(--color-red-500)` to `theme("color.red.500")` in v4. However, the reason that I use `theme()` is to get autocomplete for the tokens. (I can't memorize them all.) I don't see any autocompletion when I type `var(--color-red-500)`.

Could TailwindCSS IntelliSense be updated to provide autocomplete for the CSS variables that it generates? (Ideally, this would display the value being used on hover as well, even if the value was only a reference to another CSS variable -- depending on how Tailwind was configured. Basically the same experience as hovering a Tailwind utility class.)

---

P.S. -- Bonus points if this could also work in JavaScript files. But I'm not as concerned about that one. Maybe Tailwind could expose a `tailwindCSS.experimental.cssVarRegex` for something like that -- similar to `classRegex`.

thecrypticace:
Closes #1209

### DIFF
--- a/packages/tailwindcss-language-service/src/util/rewriting/lookup.ts
+++ b/packages/tailwindcss-language-service/src/util/rewriting/lookup.ts
@@ -8,5 +8,5 @@ export function resolveVariableValue(design: DesignSystem, name: string) {
     name = `--${name.slice(prefix.length + 3)}`
   }
 
-  return design.resolveThemeValue?.(name) ?? null
+  return design.resolveThemeValue?.(name, true) ?? null
 }
--- a/packages/tailwindcss-language-server/tests/completions/completions.test.js
+++ b/packages/tailwindcss-language-server/tests/completions/completions.test.js
@@ -2,6 +2,7 @@ import { test, expect } from 'vitest'
 import { withFixture } from '../common'
 import { css, defineTest, html, js } from '../../src/testing'
 import { createClient } from '../utils/client'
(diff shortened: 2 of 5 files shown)


packages/tailwindcss-language-service/src/util/rewriting/lookup.ts
```diff
@@ -8,5 +8,5 @@ export function resolveVariableValue(design: DesignSystem, name: string) {
     name = `--${name.slice(prefix.length + 3)}`
   }
 
-  return design.resolveThemeValue?.(name) ?? null
+  return design.resolveThemeValue?.(name, true) ?? null
```
thecrypticace, line 11: cc @philipp-spiess this correct?

thecrypticace: confirmed w/ Philipp that it is

Issue #884, opened 2023-12-17
Intellisene doesn't work in LazyVim


leoyhk:
**What version of VS Code are you using?**

LazyVim

**What version of Tailwind CSS IntelliSense are you using?**

LSP version is 0.0.14 (installed with Mason)

**What version of Tailwind CSS are you using?**

3.3.5

**What package manager are you using?**

pnpm

**What operating system are you using?**

MacOS Sonoma 14.2

**Tailwind config**

```js
import type { Config } from "tailwindcss";

export default {
  content: [""],
  theme: {
    extend: {},
  },
  plugins: [],
} satisfies Config;

```

**VS Code settings**

```json
{
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit"
  },
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.formatOnSave": true,
  "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
  "eslint.workingDirectories": [
    { "pattern": "apps/*/" },
    { "pattern": "packages/*/" }
  ],
  "typescript.tsdk": "node_modules/typescript/lib",
  "typescript.enablePromptUseWorkspaceTsdk": true,
  "tailwindCSS.experimental.configFile": "./packages/config/tailwind/index.ts",
  "[prisma]": {
    "editor.defaultFormatter": "Prisma.prisma"
  }
}

```

**Reproduction URL**
https://github.com/leoyhk/tailwind-lsp-not-working/tree/main 

**Describe your issue**
IDE: Lazyvim

LSP Version: 0.0.14 (Installed with Mason)

Bug observed: LSP does not work when "tailwindCSS.experimental.configFile": "./packages/config/tailwind/index.ts", is active in .vscode/setting.json

Expected behavior: LSP provides suggestions and auto completion

How to Recreate:

1. Open the dir with nvim (i'm using the Lazyvim distribution), open any .tsx file. LSP does not provide any TailwindCSS related suggestions when typing the classnames.
2. Enter .vscode/setting.json and comment "tailwindCSS.experimental.configFile": "./packages/config/tailwind/index.ts"
3. Restart nvim and open any .tsx file, the LSP should work normally
P.S. I'm searching for a solution / config that could make LSP work in lazyvim/nvim without changing the project settings (since it is a team project), i.e. i shall not treat commenting the line as the solution.

LspInfo shows tailwindcss is up and running, but no suggestions are given while typing
[...]

PR #1276, opened 2025-03-20
Improve error message when a workspace folder is inaccessible

thecrypticace:
On Linux when using a LSP client that does not support watching files on behalf of the server, we'll use Parcel Watcher (if possible). If we start the watcher with a non-existent or inaccessible directory, it will throw an error with a very unhelpful message: "Bad file descriptor".

The best thing we can do is an initial check for access to the directory and log a more helpful error message if it fails:

Screenshot 2025-03-20 at 12 33 29

See https://github.com/tailwindlabs/tailwindcss-intellisense/issues/884#issuecomment-1927660975

### DIFF
--- a/packages/tailwindcss-language-server/src/tw.ts
+++ b/packages/tailwindcss-language-server/src/tw.ts
@@ -186,7 +186,7 @@ export class TW {
     let base = baseUri.fsPath
 
     try {
-      await fs.access(base)
+      await fs.access(base, fs.constants.F_OK | fs.constants.R_OK)
     } catch (err) {
       console.error(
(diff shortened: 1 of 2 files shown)


packages/tailwindcss-language-server/src/tw.ts
```diff
@@ -174,6 +175,26 @@ export class TW {
   }
 
   private async _initFolder(baseUri: URI): Promise<void> {
+    // NOTE: We do this check because on Linux when using a LSP client that does
```
RobinMalfait, line 178:
```suggestion
    // NOTE: We do this check because on Linux when using an LSP client that does
```

RobinMalfait (approved):

PR #1289, opened 2025-03-27
Add CSS `var()` hovers when using v4

thecrypticace:
This adds support for hovering over things like `var(--color-red-500)` or `var(--breakpoint-xl)` and showing the value from the theme like we already support with the `theme()` function.

Additionally, I've improved the hovers for v4 theme keys. They now appear in a `@theme` block like you'd see them in your CSS and are also syntax highlighted.

Before:
Screenshot 2025-03-31 at 13 07 30

After:
Screenshot 2025-03-31 at 13 07 13


thecrypticace:
Copilot you could at least have the decency to wrap at-rule names in backticks 🤦‍♂️

philipp-spiess (approved):

PR #1292, opened 2025-04-01
v4: Improve DX around completions when prefixes are in use

### LINKED ISSUE #1291: DX issue: Using a prefix in v4 requires two actions for a completion
The quickest way for IntelliSense to support the "variant-style" prefix was to treat it as a fake variant and make some tweaks to how and what completions we show. This has an unfortunate side effect of requiring a "double action" whenever typing new classes. We should instead, when no class has been typed yet:

- Offer the prefix as a fake variant like we do today
- Show variants like normal but prefix them with the fake variant so they'll look like `tw:md:` in the suggestions
- Show classes like normal but prefix them with the fake variant so they'll look like `tw:underline` in the suggestions

None of this need apply if the variant is already typed nor if the user has typed a class name that does _not_ begin with the variant-style prefix.

thecrypticace:
Fixes #1291

- [x] Really needs tests

### DIFF
--- a/packages/tailwindcss-language-server/tests/completions/completions.test.js
+++ b/packages/tailwindcss-language-server/tests/completions/completions.test.js
@@ -741,6 +741,80 @@ defineTest({
 })
 
 defineTest({
+  name: 'v4: Completions show after a variant arbitrary value, using prefixes',
+  fs: {
+    'app.css': css`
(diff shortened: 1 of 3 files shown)


thecrypticace:
brb changing all my old projects with a prefix to use `robin:` 😂

PR #1300, opened 2025-04-07
Handle when project config is re-created

thecrypticace:
If a project's main CSS config file is deleted and then re-created the server won't re-create the project.

For now we'll restart the server when events like this happen which will ensure that the filesystem is re-scanned and the files can be picked up again.


thecrypticace:
Gonna force-merge this since it was originally part of the already approved PR #1269 — I just felt that it would be better served as its own PR

PR #1334, opened 2025-04-23
Improve language server detection for workspace folders

thecrypticace:
This PR should enable quicker booting of the language server in large multi-root workspaces

1. Potentially improved detection time in multi-root workspaces

Now we'll start the server as soon as *any* search has requested that it be started. Any remaining searches will be cancelled instead of waiting for them to complete. The language server itself will still do a more exhaustive search to find projects in all registered workspace folders.

2. Ignore pre-processor files that don't contain `@config`

Tailwind CSS v4 required us to analyze the workspace for more stylesheets that may contain `@import` (because `@import "tailwindcss"` and similar), `@theme`, etc… but these are only supported in `.css` files and not `.styl`, `.sass`, `.scss`, `.less`, etc…

We were already doing this because v3 looked for `@config` but that is valid in _any_ stylesheet type. Now we restrict `@import`, `@theme`, etc matching to `.css` files and we'll be less likely to start the server if we don't need to.

### DIFF
--- a/packages/vscode-tailwindcss/CHANGELOG.md
+++ b/packages/vscode-tailwindcss/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Ignore Python virtual env directories by default ([#1336](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1336))
 - Ignore Yarn v2+ metadata & cache directories by default ([#1336](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1336))
 - Ignore some build caches by default ([#1336](https://github.com/tailwindlabs/tailwindcss-intellisense/pull/1336))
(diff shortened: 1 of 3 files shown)


reslear:
maybe fix https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1371

solidprinciples:
@thecrypticace - facing this issue, was just delving into the codebase to understand further. 

I found that hoisting the css file to monorepo root in my case (using nuxt/ui, installed from monorepo root) fixed it.

Seems like nuxt/ui was not resolving unless placed in the appropriate folder for project detection. Bootstraping a new nuxt app, non-mono repo, was able to resolve the additional nuxt-ui styles.

Issue #1312, opened 2025-04-12
Tailwind CSS Intellisense does not work on Django project + django-tailwind


IgorArnaut:
**What version of VS Code are you using?**
v1.99.2

**What version of Tailwind CSS IntelliSense are you using?**
v0.14.15

**What version of Tailwind CSS are you using?**
v4.1.0

**What package manager are you using?**
npm

**What operating system are you using?**
Windows 11 64-bit

**Tailwind config**
```JS
export const content = [
    // Templates within theme app (e.g. base.html)
    "../templates/**/*.html",
    // Templates in other apps
    "../../**/templates/*.html",
    // Ignore files in node_modules
    "!../../**/node_modules",
];
```

**VS Code settings**

```json
{
    "[django-html]": {
        "editor.defaultFormatter": "junstyle.vscode-django-support"
    },
    "tailwindCSS.includeLanguages": {
        "plaintext": "django-html",
        "django-html": "html",
    }
}
```

**Describe your issue**
I've installed django-tailwind to my existing Django project following instructions from https://django-tailwind.readthedocs.io/en/latest/installation.html. I've also installed junstyle's Django Template Support and your Tailwind CSS Intellisense extensions to VS Code.

Folder structure:
```
└── 📁Django Real Estate Listings Web
    └── 📁.vscode
        └── settings.json
    └── 📁listings
        └── __init__.py
        └── admin.py
        └── apps.py
        └── 📁fixtures
            └── items.json
        └── forms.py
        └── 📁migrations
            └── __init__.py
            └── 0001_initial.py
            └── 0002_alter_listing_updated_at.py
        └── models.py
        └── services.py
        └── 📁templates
            └── base.html
            └── create.html
            └── detail.html
            └── index.html
            └── 📁partials
                └── __apartment.html
                └── __building.html
                └── __form.html
                └── __search.html
                └── __terms.html
            └── search.html
            └── update.html
        └── tests.py
        └── urls.py
        └── views.py
    └── 📁realestate
        └── __init__.py
        └── asgi.py
        └── settings.py
        └── urls.py
        └── wsgi.py
    └── 📁theme
        └── __init__.py
        └── apps.py
        └── 📁static
        └── 📁static_src
            └── .gitignore
            └── package-lock.json
            └── package.json
            └── postcss.config.js
            └── 📁src
                └── styles.css
            └── tailwind.config.js
            └── 📁css
    └── db.sqlite3
    └── manage.py
```

[...]

PR #1336, opened 2025-04-25
Update default file exclusions

thecrypticace:
IntelliSense, by default, ignores files for:
- Version control (`.git`, `.hg`, `.svn`)
- NPM (`node_modules`)

I'm adding three more groups to this:
- Python virtual environment folders (`.venv`, `venv`)
- Yarn v2+ metadata & caches (`.yarn`)
- *some* build caches (`.next`, `.turbo`, `.parcel-cache`, `__pycache__`)

Discovered that we should update these when triaging #1312 (that one was specifically about python virtual envs)

### DIFF
--- a/packages/tailwindcss-language-server/tests/env/ignored.test.ts
+++ b/packages/tailwindcss-language-server/tests/env/ignored.test.ts
@@ -1,0 +1,111 @@
+import { expect } from 'vitest'
+import { css, defineTest } from '../../src/testing'
+import { createClient } from '../utils/client'
+import dedent from 'dedent'
+
+let ignored = css`
+  @import 'tailwindcss';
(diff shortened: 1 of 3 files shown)


thecrypticace:
> To we want to somehow take the @source information into account in case you do want to enable intellisense in one of these folders?
> I'm thinking of the Laravel folks that use vendor (we are not excluding vendor in this case so not an issue for them. Just thinking out loud here)

For now, I'm thinking not. This list shouldn't ever contain things like `vendor`, `build`, `cache`, etc… because those could be actual names for folders that aren't caches.

This not only affects whether or not files can show hovers/completions/diagnostics but also the discovery of CSS files in the first place.

PR #1361, opened 2025-05-13
Add support for latte templates

lubomirblazekcz:
This adds support for Latte PHP templates https://latte.nette.org/en/


thecrypticace (approved):

thecrypticace:
Will be in the next release. Nore that `latte` in this case is a language ID and *not* a file extension so you'll still need an appropriate VSCode extension to provide the language to the editor (I checked and there's like 4 extensions for this and one of them seems semi-recently updated).

PR #1387, opened 2025-05-28
Properly restart server when detecting new Tailwind CSS v4 configs

thecrypticace:
In editors that use our language server directly (not our VSCode extension), if you start with a project with a CSS file that isn't a Tailwind CSS config then edit it we want the server to restart.

Unfortunately this was hitting an error causing the restart to fail but because of how things are set up this error isn't easily catchable (but it did get logged thankfully).

This fixes that error and turning a CSS file from a non-config into a proper Tailwind CSS v4 config should work and the server should react accordingly.


philipp-spiess (approved):
Small enough change 👍

PR #1396, opened 2025-06-02
Fix high CPU usage when encountering non-file URI workspaces

### LINKED ISSUE #1394: High CPU usage with virtual files in neovim

N/A

v0.14.20

v4.0.6

pnpm

Linux

**Tailwind config**

Unrelated

**VS Code settings**

N/A

**Reproduction URL**

Any project should trigger this issue. I can provide neovim-specific steps if necessary

**EDIT**: minimal steps: (0) Obviously, configure the tailwind LSP
1. In a folder that matches tailwind's `root_dir` (i.e., any folder with tailwind project should suffice), launch nvim
2. Create and open a new "virtual file" with `:edit foo://bar.svelte`

**Describe your issue**

Recently, neovim (0.11) introduced a new way to configure LSPs. Unfortunately, this new method changed the conditions of when a server attaches to files: by default, LSPs are attached even to "virtual files" (i.e., files that do not exist on disk; these usually come from extensions). When this happens, the tailwind LSP gets stuck with 100% CPU usage until killed.

thecrypticace:
Fixes #1394

There are two issues here:
- Non-file schemes were being treated as if they were `file://` URIs and we were getting their filesystem path when ends up being an empty string.
- Neovim, when opening a virtual file (e.g. `foo://bar.html`), causes the server to be initialized with a file URI of `file://.` and rootPath of `/.`. I'd expect them to use the `foo://` scheme but that does not happen. I think it's super reasonable to have the server refuse a workspace folder that _is_ the filesystem root. This check is definitely not exhaustive (may not handle network shares, arbitrary mounts, different drives on Windows, etc…) but this is probably "good enough" to detect this situation. 

Both of these results in the folder being treated as `/` and results in a file search from the root causing super high CPU usage. I've also added a check inside `getProject` so files with non-file schemes are ignored. They'll get a `fsPath` of empty string as well and as such won't match any project. The VSCode extension already does this at the extension level but we weren't doing it in the language server.

### DIFF
--- a/packages/tailwindcss-language-server/src/tw.ts
+++ b/packages/tailwindcss-language-server/src/tw.ts
@@ -122,14 +122,23 @@ export class TW {
     await this.initPromise
   }
 
+  private validateFolderUri(uri: URI): boolean {
+    if (uri.scheme !== 'file') {
+      console.warn(
+        `The workspace folder [${uri.toString()}] will be ignored: it does not use the file scheme.`,
+      )
+      return false
+    }
+
+    return true
+  }
+
   private getWorkspaceFolders(): WorkspaceFolder[] {
     if (this.initializeParams.workspaceFolders?.length) {
       return this.initializeParams.workspaceFolders.flatMap((folder) => {
         let uri = URI.parse(folder.uri)
-        if (uri.scheme !== 'file') {
-          console.warn(`Non-file workspace folder will be ignored: ${folder.uri}`)
-          return []
-        }
+
+        if (!this.validateFolderUri(uri)) return []
 
         return [
           {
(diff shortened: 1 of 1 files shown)


igorlfs:
Hey, 

> There's three things wrong here:
>  * `rootPath` is `/.` which is wrong. For a non-file url I'm not sure that there is _any_ valid value for this property.
>  * `rootUri` should _probably_ be `foo://.`
>  * `workspaceFolders.0.url` should  _probably_ be `foo://.`
> 
> 
> I think it's probably wrong for it to be `file://`. Ultimately this resolves to `/` according to `vscode-uri`:

Thanks for the info. I'll open an issue in neovim.

PR #1401, opened 2025-06-04
Improve stylesheet "root" detection

### LINKED ISSUE #1388: intellisense not works after v0.14.16

Version: 1.100.2

For example: v0.14.19

v^4.0.7
tried on a newer version same bug

npm

macOS

**Tailwind config**

```js
@import 'tailwindcss/theme.css' layer(something) prefix(sm) important;
@layer something{
  .test{
    @import 'tailwindcss/preflight.css';
    @tailwind utilities;

    font-family: var(--sm-font-main);
  }
}
```

**VS Code settings**

```json
{
  "files.associations": {
    "*.css": "tailwindcss"
  },
}
```

**Reproduction URL**

Private repo

**Describe your issue**

The problem is that after v0.14.16 extension version autocompletion stopped worked
No errors in tailwind CSS IntelliSens ouptut
Before:
Image

After:
Image

thecrypticace:
Fixes #1388

There's two problems here:
- We consider any stylesheet with an `@import` a potential root. This includes *compiled* stylesheets that have URL imports to things like google fonts
- We consider all "roots" graphs equally instead of prioritizing ones that import Tailwind CSS.

### DIFF
--- a/packages/tailwindcss-language-server/src/project-locator.test.ts
+++ b/packages/tailwindcss-language-server/src/project-locator.test.ts
@@ -499,6 +499,38 @@ testLocator({
   ],
 })
 
+testLocator({
+  name: 'Stylesheets that import Tailwind CSS are picked over ones that dont',
+  fs: {
+    'a/foo.css': css`
+      @import './bar.css';
+      .a {
+        color: red;
+      }
+    `,
+    'a/bar.css': css`
+      .b {
+        color: red;
+      }
+    `,
+    'src/app.css': css`
+      @import 'tailwindcss';
+    `,
+  },
+  expected: [
+    {
+      version: '4.1.1 (bundled)',
+      config: '/src/app.css',
+      content: [],
+    },
+    {
+      version: '4.1.1 (bundled)',
+      config: '/a/foo.css',
+      content: [],
+    },
+  ],
+})
+
 // ---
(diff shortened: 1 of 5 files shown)


packages/tailwindcss-language-server/src/project-locator.test.ts
```diff
@@ -124,11 +124,15 @@ testFixture('v4/workspaces', [
...
       '{URL}/packages/admin/tw.css',
+      '{URL}/packages/shared/**',
+      '{URL}/packages/shared/ui.css',
     ],
   },
   {
     config: 'packages/web/app.css',
     selectors: [
+      '{URL}/packages/shared/**',
+      '{URL}/packages/shared/ui.css',
```
thecrypticace, line 135: I think these showing up indicates a problem with this approach. These show up because this is an imported file. *this is generally fine*

However, in this case it might be unexpected. Especially if `packages/shared` has its own Tailwind CSS config. It'll prefer packages/admin over packages/shared and that feels like a bug. 🤔

thecrypticace: I figured out a temporary solution that doesn't introduce potential issues like this. But it's kinda awful.

Will ship with it but it could use some improvement later.

PR #1409, opened 2025-06-10
Fix variable and comment highlighting in `@theme`

thecrypticace:
Fixes #1365
Fixes #1402

PR #1410, opened 2025-06-11
maybe fix windows ci failures

thecrypticace:
Noticed that Windows CI just suddenly started failing in #1409 and have no idea why. All tests pass locally on Windows and do so reasonably quickly so I guess I have to triage this directly on CI 😭


thecrypticace:
It's literally because Windows I/O got like 10x slower. Everything is cloned and installed on the C: drive now instead of D: and pnpm install takes over a minute where it took 7–10s before.

Great…

PR #1425, opened 2025-07-27
Add custom css class from config files to completion

felix-droese:


felix-droese:
@thecrypticace 

thanks for your insight.

How would you benchmark speed?
Should I proceed with this PR and approach or should we close it?

PR #1428, opened 2025-07-31
Match class functions that appear after an opening square bracket

javoscript:
**The problem:**
The lookbehind pattern `(?<=^|[:=,;\s{()])` doesn't include `[` in its character class, so these cases would be missed:

```javascript
const classes = [clsx('btn', 'primary')]  // ❌ "clsx" not matched
const arr = [cn('text-red-500')]          // ❌ "cn" not matched  
```

But these would work:

```javascript
const classes = [ clsx('btn', 'primary')] // ✅ matches (whitespace after [)
const arr = [, cn('text-red-500')]        // ✅ matches (comma before cn)
```

**Why this matters:**
Array contexts are common in modern JavaScript/React for combining classes:

```javascript
const buttonClasses = [clsx(conditionalClasses), baseClasses]
const styles = [cn(variantStyle), defaultStyle]
```

**The Fix:**
The regex should include `[` in the lookbehind character class:

```javascript
/(?<=^|[:=,;\s{()\[])([\p{ID_Start}$_][\p{ID_Continue}$_.]*)[(`]/dgiu
//               ^ add \[ here
```

This would ensure we properly detect class utility functions at the beginning of arrays.

### DIFF
--- a/packages/tailwindcss-language-service/src/util/find.test.ts
+++ b/packages/tailwindcss-language-service/src/util/find.test.ts
@@ -1081,3 +1081,30 @@ test('class functions work inside astro code fences', async ({ expect }) => {
     },
   ])
 })
+
+test('classFunctions are detected inside of arrays in javascript just after opening bracket', async ({ expect }) => {
+  let file = createDocument({
+    name: 'file.js',
+    lang: 'javascript',
+    settings: {
+      tailwindCSS: {
+        classFunctions: ['cn'],
+      },
+    },
+    content: js`
+      const list = [cn('relative')]
+    `,
+  })
+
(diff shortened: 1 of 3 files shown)


javoscript:
**Examples:**

Not working without the space before the bracket:
Screenshot 2025-07-31 at 20 23 10

Working with the space:
Screenshot 2025-07-31 at 20 23 41

thecrypticace (approved):

thecrypticace:
I'd say I can't believe I missed that but… I can. 😅

Thanks! I'll tag a release with the fix tomorrow 👍

PR #1441, opened 2025-08-29
WIP: Various Performance Improvements

joshuaepstein:
- Add caching to workspace scanning
- Add selecting watching, pattern limits to file watching
   - Disabled unnecessary chokidar features (alwaysStat, usePolling)
- Added configurable debounce time and adjustable debounce time via settings to rebuilds (could lead to CPU spikes)
- Added 5 minute intervals for inactive project pruning, enhanced language model cache, added force cleanup, 10 minute inactivity before projet removal

### DIFF
--- a/packages/tailwindcss-language-server/src/language/languageModelCache.ts
+++ b/packages/tailwindcss-language-server/src/language/languageModelCache.ts
@@ -9,6 +9,7 @@ export interface LanguageModelCache<T> {
   get(document: TextDocument): T
   onDocumentRemoved(document: TextDocument): void
   dispose(): void
+  forceCleanup(): void
 }
 
 export function getLanguageModelCache<T>(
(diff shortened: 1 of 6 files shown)


thecrypticace:
I am curious — did you use AI to write this PR? What was the reasoning behind all of these changes? Can you point to actual problems caused in your own projects? If you can you can open an issue there detailing them with a reproduction and we can start there.

1. Calls to `getConfiguration()` *already* hit an internal cache. They don't need to be debounced.
2. `pruneLanguageModelCaches` does nothing but log (also this cleanup already happens)
3. `getProjectLastActivity` *does nothing* in practice. It always considers all projects active.
4. Disabling polling means this doesn't work in docker on anything other than Linux
5. Limiting the watch patterns without any sort of analysis of the patterns means it may miss files
6. `disableCodeLens` and `disableColorDecorators` options are unncessary and already have options that exist that the user can use to disable them called `tailwindCSS.codeLens` and `tailwindCSS.colorDecorators`.
7. You bumped the version number (generally don't do this for projects you don't own — this is the job of a maintainer)
8. There are zero change log entries for this nor readme entries regarding any newly introduced options

I appreciate the sentiment behind the idea of "improve performance" but this isn't how to approach this.

PR #1450, opened 2025-09-04
Support elixir and phoenix heex files in neovim

patryk-s:
In neovim, the newer Phoenix HTML templates are [identified as `heex` files](https://github.com/neovim/neovim/blob/2b421d518faec9e1b0a6d1c46026d2f13f4fb73f/runtime/lua/vim/filetype.lua#L571). Additionally, there can be HTML code within `.ex` files (via `~H""`), which both neovim and VSCode identify as `elixir`.


thecrypticace:
I think treating Elixir as HTML *might* be a bit finicky (ideally we'd detect the code inside ~H"…") but will handle that later if it proves to be an issue.

thecrypticace:
Will be updated in the next release of the language server.

Thr3x:
Does this mean Elixir wasn't even supported? I've been pulling my hair out trying to get intellisense to work with v4 tailwind in elixir files....

thecrypticace:
> Does this mean Elixir wasn't even supported? I've been pulling my hair out trying to get intellisense to work with tailwind v4 in elixir files.... Elixir files worked just fine being detected as HTML with Tailwind v3. @thecrypticace

It was but it depends on the language ID sent to our language server: https://github.com/tailwindlabs/tailwindcss-intellisense/blame/ede2c7d0fb2f0edc8f2cd17bd8c0ca0da1451e35/packages/tailwindcss-language-service/src/util/languages.ts#L21-L23

Note: language IDs are **not** file extensions. They're independent.

Those Elixir languages have been there for a little over 5 years — if the language ID doesn't match or otherwise changed to something else then it won't work.

You can always work around this by using our [`tailwindCSS.includeLanguages`](https://github.com/tailwindlabs/tailwindcss-intellisense?tab=readme-ov-file#tailwindcssincludelanguages) option which can map arbitrary language IDs to one we recognize.

PR #1475, opened 2025-10-06
Add canonical class diagnostics

thecrypticace:
This feature is *off* by default due to performance and memory considerations. It's also possible that this won't even be the final version of this.

We've internally discussed linting per class *list* instead of per class along with some other stuff but getting this here to test.

Tests won't pass in CI until we have at least an insiders build to test with.


pondorasti:
are there any plans for a codemod to autofix canonical warns? 

we have a large monorepo with upwards of hundreds of files that could be automatically cleaned up with the help of a codemod

PR #1508, opened 2025-12-02
Add command to apply all canonical class suggestions at once (#2)

damien-schneider:
Added a command to batch apply all the suggested class changes.


thecrypticace:
This should probably be implemented as a "fix all" code action and not a custom command.

See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction

Look for `SourceFixAll`

Issue #1417, opened 2025-06-26
Tagged Templates don't have hover preview in Vue components


cooper667:
**What version of VS Code are you using?**

Version: 1.101.2

**What version of Tailwind CSS IntelliSense are you using?**

Version: 0.14.23

**What version of Tailwind CSS are you using?**

4.1.8

**What package manager are you using?**

Yarn

**What operating system are you using?**

MacOs

**Tailwind CSS Stylesheet (v4) or config file (v3)**

v4

**VS Code settings**

```json
  "tailwindCSS.classFunctions": ["tw"],
  "tailwindCSS.emmetCompletions": true,
  "tailwindcss-intellisense.trace.server": "verbose",
  "tailwindCSS.experimental.configFile": "./pkg/xxx/src/index.css"
```

**Describe your issue**

Given a fn `const tw = (strings: TemplateStringsArray, ...values: unknown[]) =>
  String.raw({ raw: strings }, ...values);`

The following give hover information in TS files, but not in the `<script lang="ts">` part of .vue files

```
const defaultClasses = {
  field: tw`relative`,
  fieldInput: tw`bg-neutral-50 peer to-ba absolute top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2 cursor-pointer appearance-none rounded border transition disabled:cursor-default disabled:opacity-50`,
};
const x = tw`border-l`;
```

thecrypticace:
Ah it works if you put the script before the template but not after. Yep okay.

Minimal repro:
```html
<template></template>
<script>
let a = tw`absolute`;
</script>
```

RobinMalfait:
This should be fixed by #1580 and will be available in the next release. It will be available in the pre-release version shortly.

PR #1529, opened 2025-12-31
Simplify document class list detection

thecrypticace:
This should be mostly functionally equivalent. It simplifies a good chunk of the code used throughout to find class lists and class names.

1. It eliminates many repeated calls to `getConfiguration(…)` which would increase based on the number of embedded languages in the document. These hit a cache anyway so the overhead wasn't significant but the repeated calls aren't necessary.
2. Because of this I've now made several things synchronous where they were async before.
3. It eliminates a search range limitation for hovers which would cause some class name hovers to fail on large enough documents depending on document structure and what came *after* the class name further in the document.
4. This will simplify a future refactor that caches language boundaries, class lists, and class names on a per-document basis and only recompute when documents change.

### DIFF
--- a/packages/tailwindcss-language-service/src/diagnostics/canonical-classes.ts
+++ b/packages/tailwindcss-language-service/src/diagnostics/canonical-classes.ts
@@ -3,11 +3,11 @@ import type { State, Settings } from '../util/state'
 import { type SuggestCanonicalClassesDiagnostic, DiagnosticKind } from './types'
 import { findClassListsInDocument, getClassNamesInClassList } from '../util/find'
 
-export async function getSuggestCanonicalClassesDiagnostics(
+export function getSuggestCanonicalClassesDiagnostics(
   state: State,
   document: TextDocument,
   settings: Settings,
-): Promise<SuggestCanonicalClassesDiagnostic[]> {
+): SuggestCanonicalClassesDiagnostic[] {
   if (!state.v4) return []
   if (!state.designSystem.canonicalizeCandidates) return []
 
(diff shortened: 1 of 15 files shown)


Thy3634:
This refactor would also fix a real user-facing bug worth tracking here: **lint diagnostics don't run on `classFunctions` class lists in the `<script>` block of `.vue` SFCs**.

Repro (v0.16.0, Tailwind v4, `tailwindCSS.classFunctions: ["cn"]`):

```vue
<template>
  <div class="[color:red]"></div> <!-- suggestCanonicalClasses fires -->
</template>
<script setup>
const x = cn("[color:red]") // no diagnostic at all
</script>
```

Completions and hover work in the script, so only lint is affected — easy to miss.

Root cause on `main`: the diagnostics entry `findClassListsInDocument()` only forwards `html`/`jsx` and `css` boundaries to the class list extractors; `js` boundaries are never processed. `findClassListsInHtmlRange()` does scan `js` boundaries for `classFunctions`, but it is invoked with the template boundary's range and extracts text from the template-only string, so the script matches resolve to empty text and get dropped. Completions/hover go through `provideClassAttributeCompletions` / `findClassNameAtPosition`, which map `js` boundaries to `findClassListsInRange(…, 'jsx', range)`, so they work. #1417 / #1580 fixed a similar offset bug in the hover path; the lint path is still broken.

The new `findClassListsInDocument()` routes `js`/`jsx` boundaries through `findClassListsInHtmlRange()` with their own ranges, which resolves this. A regression test asserting that class lists from a `<script>` block are returned for an SFC would pin it down.

Workaround I'm using in the meantime (full-document scan path):

```json
"tailwindCSS.experimental.classRegex": [
  ["(?<![\\w$.])cn\\(((?:[^)(]*|\\([^)(]*\\))*)\\)", "[\"'\\`]([^\"'\\`]*)[\"'\\`]"]
]
```

PR #1571, opened 2026-04-21
Add canonical class fix-all actions for files and workspaces

TurtIeSocks:
## Summary

Adds bulk-fix support for Tailwind canonical class suggestions in two places:

1. a standard LSP fix-all code action for the current file
2. a VS Code command to apply canonical class suggestions across the current workspace folder

### File-level fix all
Before:
- users only saw one warning fix at a time, like `Replace with 'print:flex'`

After:
- users still get the single-instance quick fixes
- users also get a standard Tailwind fix-all code action for the current file:
  - `Fix all Tailwind CSS canonical class suggestions`

### Workspace-level fix all
Before:
- there was no built-in way to apply canonical class suggestions across multiple files in the workspace

After:
- there is a new command:
  - `Tailwind CSS: Fix Canonical Class Suggestions in Workspace`
- it scans the current workspace folder and applies canonical class suggestions across all matching files
- it skips Git-ignored files

## Implementation details

- added `source.fixAll.tailwindcss` support to the language server code actions
- kept the existing per-warning quick fixes intact
- advertised Tailwind fix-all support through server capabilities
- added a VS Code command that requests per-file Tailwind fix-all actions from the language server and merges them into one workspace edit
- filtered workspace scans to ignore Git-ignored files

## Examples

Before:
```html
<div class="[@media_print]:flex [color:red]/50 mt-[16px]"></div>
```
After:
```html
<div class="print:flex text-[red]/50 mt-4"></div>
```

In a single file:
- the new fix-all action applies all canonical replacements in that document

Across a workspace:
- the new command applies the same kind of fixes across all non-gitignored files in the current workspace folder

## Testing
- added language-server tests covering the new file-level fix-all action
- added capability coverage for advertised Tailwind fix-all support
- rebuilt @tailwindcss/language-service
- rebuilt @tailwindcss/language-server
- rebuilt vscode-tailwindcss

## Notes
- the file-level behavior follows the review feedback on the earlier PR by using a standard LSP fix-all action instead of a custom command
- workspace-wide fixing is implemented as a separate VS Code command because LSP textDocument/codeAction is document-scoped

Supersedes #1508

### DIFF
--- a/packages/tailwindcss-language-server/src/tw.ts
+++ b/packages/tailwindcss-language-server/src/tw.ts
@@ -80,6 +80,8 @@ const TRIGGER_CHARACTERS = [
   '-',
 ] as const
 
+const CODE_ACTION_KINDS = ['quickfix', 'source.fixAll.tailwindcss']
+
 async function getConfigFileFromCssFile(cssFile: string): Promise<string | null> {
   let css = await readCssFile(cssFile)
   if (!css) return null
@@ -925,7 +927,10 @@ export class TW {
     }
 
     if (client.textDocument?.codeAction?.dynamicRegistration) {
-      capabilities.add(CodeActionRequest.type, { documentSelector: null })
+      capabilities.add(CodeActionRequest.type, {
+        documentSelector: null,
+        codeActionKinds: CODE_ACTION_KINDS,
+      })
     }
 
     if (client.textDocument?.codeLens?.dynamicRegistration) {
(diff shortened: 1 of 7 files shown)


TurtIeSocks:
Bump?

Jasperhuting:
I reproduced a workspace-only issue in a real Tailwind CSS 4.3.3 project: the command reported no canonical suggestions while more than 500 remained. The file-level fix worked.

The workspace loop opens a document and immediately requests `textDocument/codeAction`, but the language server returns no actions until that document is present in its `DocumentService`. For files that were not already open, synchronization may not have completed yet.

I prepared a focused draft PR against your `fix-all-code-action` branch: https://github.com/TurtIeSocks/tailwindcss-intellisense/pull/1

It keeps the existing file-level source action unchanged, sends document content through a dedicated workspace request, and adds a regression test for a document that is not open in the LSP. The code-action tests and VS Code extension typecheck pass, and the packaged extension was manually verified against the original workspace.

PR #1578, opened 2026-05-06
Mark `@variant` as deprecated when defining custom variants

RobinMalfait:
This PR adds a diagnostic for when you are usint `@variant` for creating custom variants.
When creating custom variants, `@custom-variant` should be used instead.

I was struggling a little bit with the name of the diagnostic, because `@variant` on its own is not deprecated. It's only deprecated when using it for creating custom variants.

Additionally, naming the diagnostic related to `deprecateCustomVariantDirective` or similar makes it seem like `@custom-variant` is deprecated but that is what we _should_ use. Instead, I opted for a name where we mark the `variant` _syntax_ as deprecated.

Definitely not perfect, open for suggestions but I think the naming is fine.

The deprecation message itself looks like this:
> `@variant` is deprecated for defining custom variants. Use `@custom-variant` instead.

## Test plan

Shows a warning by default with a suggestion attached:
image
image


packages/tailwindcss-language-service/src/diagnostics/getDeprecatedAtRuleDiagnostics.ts
```diff
@@ -0,0 +1,62 @@
...
+          },
+          range,
+        ),
+        severity:
+          severity === 'error'
+            ? 1 /* DiagnosticSeverity.Error */
+            : 2 /* DiagnosticSeverity.Warning */,
+        message:
+          '`@variant` is deprecated for defining custom variants. Use `@custom-variant` instead.',
+        suggestions: ['@custom-variant'],
```
thecrypticace, line 56: Should add a deprecated diagnostic tag to these. Probably need to check client capabilities to make sure tags are supported first though (they appeared in protocol v3.15)

RobinMalfait: Not a bad idea! Now it renders as deprecated: 
image

PR #1580, opened 2026-05-12
Fix class function hover lookup in Vue script blocks

### LINKED ISSUE #1417: Tagged Templates don't have hover preview in Vue components

waradu:
With this hover now works after the template:

image

Matches were using absolute document offsets, but the parser expects range-relative offsets, so hover lookup could miss classes. This should hopefully also fix #1417.

I also added a small regression test.

### DIFF
--- a/packages/tailwindcss-language-service/src/util/find.ts
+++ b/packages/tailwindcss-language-service/src/util/find.ts
@@ -205,6 +205,7 @@ export async function findClassListsInHtmlRange(
   const matches = matchClassAttributes(text, settings.classAttributes)
 
   let boundaries = getLanguageBoundaries(state, doc)
+  let rangeStartOffset = range ? doc.offsetAt(range.start) : 0
 
   for (let boundary of boundaries ?? []) {
     let isJsContext = boundary.type === 'js' || boundary.type === 'jsx'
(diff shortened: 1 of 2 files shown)


packages/tailwindcss-language-service/src/util/find.ts
```diff
@@ -205,21 +205,22 @@ export async function findClassListsInHtmlRange(
...
-    let offset = doc.offsetAt(boundary.range.start)
+    let offset = doc.offsetAt(boundary.range.start) - rangeStartOffset
     let fnMatches = matchClassFunctions(str, settings.classFunctions)
 
     fnMatches.forEach((match) => {
       if (match.index) match.index += offset
     })
 
-    matches.push(...fnMatches)
+    matches.push(...fnMatches.filter((match) => match.index >= 0 && match.index < text.length))
```
RobinMalfait, line 223: I don't think this part is necessary. Or at least, I can't figure out a test that requires this additional filter.

RobinMalfait (approved):
Thanks for the PR!


Looks like I can't push to your branch for some reason:

Image

Will revert that change on the `main` branch, thanks!

